PR #12, opened 2020-11-26
Moves Aegis Hardsuits to Secure Armory

Michiyamenotehifunana:
## About The Pull Request

Aegis' Hardsuits are now no longer just sitting in the regular armory, they are now locked behind the secure armory's glass.

dreammaker_3A0fGwUOZU

Hardsuit modules still spawn at the same place, but now there is only a total of 6 spawns instead of 8.

dreammaker_Y8cOkuq8XF

## Changelog
```changelog
add: Aegis Hardsuits are now locked up in the secure armory.
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -1315,13 +1315,6 @@
 /obj/item/weapon/cell/small/high,
 /turf/simulated/floor/tiled/dark/cargo,
 /area/eris/security/armory)
-"adk" = (
-/obj/structure/table/rack,
-/obj/spawner/rig_module,
-/obj/spawner/rig_module,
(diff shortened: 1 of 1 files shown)


Jamini:
I thought we were removing one, moving one to the secure armory, and putting one in the Commander's room?

ShadzFen (changes requested):
Rather than 6 changes in the middle of the armoury, have 1 100% chance of spawning in the armoury, and 1 100% chance of spawning in the commander's room or wherever Jams wanted it to spawn.

If this turns out to be aids for gameplay, we can always change it later.

Michiyamenotehifunana:
In the interest of keeping our map maintainable, I shall be closing this PR as these changes were made in FastDMM which really fucks up the tile dictionary. I shall redo it with the changes in mind in StrongDMM.

PR #13, opened 2020-11-26
Upstream Patch to make circular saws and reinforced platings not overly common

Michiyamenotehifunana:
## About The Pull Request

Exactly what it says on the tin. Ports https://github.com/discordia-space/CEV-Eris/pull/5542

## Changelog
```changelog Toriate
tweak: Ported an upstream patch that should reduce the incidence of circular saw and reinforced plating spawns.
```


ShadzFen (approved):
It also adds spawntags to a lot of tools. Looks fine, I guess. The variables have been removed rather than modified, so I assume they're just set to the average spawnrate.

PR #30, opened 2020-11-29
Fixes hivemind announcement delay

Michiyamenotehifunana:
## About The Pull Request

This PR will finally fix the wonky announcement behavior of hiveminds. Instead of immediately giving away their location, it hooks onto the same delayed level 7 biohazard announcement thing and calls out its location 300 seconds after it is spawned. This should give the hivemind enough time to establish itself as a threat.

## Changelog
```changelog Toriate
fix: fixed the hivemind's location callout announcement being instant
```

### DIFF
--- a/code/game/gamemodes/events/hivemind_invasion.dm
+++ b/code/game/gamemodes/events/hivemind_invasion.dm
@@ -17,14 +17,15 @@
 
 /datum/event/hivemind
 	announceWhen	= 300
+	var/turf/start_location	// OCCULUS EDIT - moved from start()
 
 
 /datum/event/hivemind/announce()
 	level_seven_announcement()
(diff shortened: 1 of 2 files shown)


ShadzFen:
I don't think upping the number of players it requires to spawn is necessary. I think we just need it to be rarer or flat out not spawn unless there's 9 or more people **playing**.

Michiyamenotehifunana:
req_crew changes no longer relevant, PR is now atomic

Jamini:
I still feel we shouldn't announce where it spawns. But delaying the message seems good.

Jamini (approved):
Adds a delay on this even. Nothing too exceptional here. We still probably shouldn't annouce exactly where it is.

PR #31, opened 2020-11-29
adjusted blacklists 

Jamini:
Blacklisted a bunch of stuff that shouldn't be spawning, kept their child objects that should be spawning

## About The Pull Request

Fixes some blacklists

## Why It's Good For The Game

Makes stuff that shouldn't be spawning, not spawn
Added security network to the Aegis rig

## Changelog
```changelog
fix: item blacklists
fix: added security network to Aegis hardsuit helmet.
```


ShadzFen (approved):
Modular blacklist items. All good.

PR #34, opened 2020-11-30
Map fixes: aft section 2 blastdoor and shower airlock names

Michiyamenotehifunana:
## About The Pull Request

This PR will fix the opacity of the aft section 2 blastdoor, and also fixes the airlock names on the auxiliary closet (the old showers) and the changing room (the new showers).

## Changelog
```changelog Toriate
fix: Aft Section 2 Blastdoor is no longer wonky
fix: Aux closet and showers airlocks are now appropriately named
```


Jamini (approved):
Nothing unusual here. Looks good Tori!

PR #47, opened 2020-12-01
Adds a disk of civilian grade guns to the ammo vendors

Michiyamenotehifunana:
## About The Pull Request

pewpewPreview

Featuring a pistol caliber carbine, crappy """SMG""", and integrally silenced pistol, this new pack of guns adds more varieties of civilian tier guns. The guns can either be found in maint, or be printed from a disk that can be purchased from the ammo vendors.

The PCC has slightly higher stopping power than the C-20, but far less penetrative capabilities.
The """SMG""" cannot fire in full auto, and has very average stats.
The pistol is just about in line with the Mk58 damage-wise, but with far less penetration.

## Why It's Good For The Game

dreamseeker_IYvIOfFMKq

More guns are fun! Especially guns you can print out yourself!

## Changelog
```changelog Toriate
add: A cache of long lost gun designs has been uncovered and put back into production by various small-scale manufacturers. Find a convenient compilation of the designs at your nearest ammo vendor!
```


Jamini (changes requested):
Could you put up an emojicracy on these before we merge it Tori? I don't object to the guns, but we should get feedback from the crew before adding them.

Jamini (approved):

PR #49, opened 2020-12-01
The game will not start without anyone ready

### LINKED ISSUE #39: The round is capable of starting with no players readied up
#### Description of issue

Ideally the round should not start if no players are ready to play

Currently the round starts as long as there is a player in the lobby.

#### Steps to reproduce 

Join the server and don't ready up.

Very likely an eris-specific change we need to modify. Most SS13 servers do not start if nobody is ready to play.

#### Issue bingo

cebutris:
## About The Pull Request

Prevents the game from starting without a player ready. Given a full 5 minutes of testing to ensure it works properly. 

The only issues I can see with this is: 
Some people (me) like to start as a drone to prepare for solo rounds by setting the engine. This will make that impossible.
If a player does not have a role properly set, they will not count for the check (though in this case, usually they would be returned to lobby at startup anyway)

## Why It's Good For The Game

Empty station sad

## Changelog
```changelog Cebutris
tweak: Game will not start if nobody is ready
```

Fixes #39

### DIFF
--- a/code/game/gamemodes/storyteller.dm
+++ b/code/game/gamemodes/storyteller.dm
@@ -78,8 +78,10 @@ GLOBAL_DATUM(storyteller, /datum/storyteller)
 
 	var/engineer = FALSE
 	var/command = FALSE
+	var/readyplayer = FALSE//SYZ edit, make sure that the round doesn't start if nobody is ready
 	for(var/mob/new_player/player in GLOB.player_list)
 		if(player.ready && player.mind)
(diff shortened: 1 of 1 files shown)


ShadzFen:
This PR fixed #39 and you should label it as such so that github closes that ticket once it merges.

ShadzFen (approved):
Looks fine, non-modular code, though this can't be made modular so. Good to go.

PR #50, opened 2020-12-02
adds a mcmaint bar

bearrrrrrrr:
## About The Pull Request

adds a maint bar. maint bar has private room w/ bed. maint bar buried behind some trash.
image

## Why It's Good For The Game

maint bar.

## Changelog
```changelog
add: adds maint bar
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -34419,14 +34419,14 @@
 /obj/effect/floor_decal/industrial/warning{
 	dir = 1
 	},
-/obj/spawner/flora/low_chance,
+/obj/spawner/scrap/dense/low_chance,
 /turf/simulated/floor/tiled/techmaint,
 /area/eris/maintenance/section4deck4central)
 "bCU" = (
-/obj/machinery/portable_atmospherics/powered/pump,
 /obj/effect/floor_decal/industrial/warning{
 	dir = 1
 	},
+/obj/spawner/scrap/dense,
 /turf/simulated/floor/tiled/techmaint,
 /area/eris/maintenance/section4deck4central)
 "bCV" = (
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
im bad at writing pr's

ShadzFen:
Well some more details would be nice. What floor is this on? What side of the ship and position is it exactly in? Does it replace another room? If so, what did that room contain before? If it adds onto the ship, have you tested that shields don't go through half the room?

bearrrrrrrr:
> Well some more details would be nice. What floor is this on? What side of the ship and position is it exactly in? Does it replace another room? If so, what did that room contain before? If it adds onto the ship, have you tested that shields don't go through half the room?

in order: Deck four section three central maintenance, doesn't replace anything, shields do not go through it. I tested stuff before I pushed a PR, for once, which is a really weird feeling lol
image

Jamini:
A major change like this probably should get an emojicracy on it. Since it's a fairly big add.
I'm also not really convinced we *need* a maint bar. Our population isn't really high enough to support the club, much less competition.

bearrrrrrrr:
The bar doesn't have any machinery, or any glasses, and completely random liquor along with a bunch of trash. It's a loot area, not really a bar area. You can't feasibly set up shop. It lacks consistent spawns. The only thing guarenteed to spawn is a small pile of cash at about 100 credits.

Jamini (approved):
Emojicracy passed. doesn't look objectionable. I still *personally* object to this PR, but in this case the community has stated their desire. LET THE FUCKROOMS BEGIN!

bearrrrrrrr:
I'll significantly nerf loot or just change it to another Garbage Room:tm: if people consistently use it for a maintbar :ok_hand:

ShadzFen:
Appreciated. But you will be held to your word 

image

PR #52, opened 2020-12-02
Fixes Job Department Constants

Michiyamenotehifunana:
## About The Pull Request

This effectively causes the Merchant and First Officer to once again count as heads, when previously they did not. An issue with the First Officer's spawner has also been fixed.

Furthermore, this allows [REDACTED] to work properly for the First Officer and Chaplain positions again.

## Changelog
```changelog Toriate
fix: Job department constants have been fixed and merchants should now count as command staff again.
```

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -1,15 +1,14 @@
 #define ASSISTANT_TITLE "Deckhand"
 
 //Jobs depatment lists for use in constant expressions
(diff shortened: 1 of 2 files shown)


ShadzFen (approved):
All good, non-modular. For future reference, the administrative tag is for code changes/fixes regarding admin functions. That means admin verbs and what not. Not actual administrative work IG.

PR #53, opened 2020-12-02
Secret notes added to maint drops [Spoiler warning]

Ikmachek:
## About The Pull Request
I had a wicked idea, since every child of /paper spawns in maint loot, why not add some sinister history to the ship with this method. Reading this PR does have some mild spoilers so unless you're reviewing it, you might not wanna look too hard. Tested to make sure it compiles and the format comes out fine, now if it spawns in maint I didn't have the time to check, here's hoping.

## Why It's Good For The Game
Because the ship has a history, despite [redacted] trying to hide it.

## Changelog
```
add: Some lost letters, notes and emails to maint loot.
```


Jamini:
Okay this should be fixed now. :angry:

Jamini (approved):
Modular lore drops for maints. Sekrets! Everything looks good on this end from me.... bluuugh

PR #74, opened 2020-12-11
Ikky's first map PR

Ikmachek:
## About The Pull Request
For months people complained we have no easy to read regs, so I fixed that. Now I mapped it,  and fixed the barracks being Ironhammer

## Why It's Good For The Game
So now security has a book of why you're in the brig for 45 minutes.

## Changelog
```changelog
add: Aegis Regulations book to the map.
```


Michiyamenotehifunana (approved):
This seems to even rename the security post airlocks, but eh!

PR #75, opened 2020-12-12
tiny optimization that probably does nothing

Michiyamenotehifunana:
## About The Pull Request
This PR will remove some crappy copypasta and timers with the aim of reducing character generation lag. This does not appear to have an impact on 500+ body marking characters, but it should help a bit. Juuuuust a little bit.

No noticeable issues were noticed from the removal of the spaghetti, and even in the worst case, it'd just render character previews unavailable until you refresh the character creator screen.

## Changelog
```changelog Toriate
tweak: made a teeny tiny character generation microoptimization that no one will notice
```

### DIFF
--- a/code/modules/client/preference_setup/preference_setup.dm
+++ b/code/modules/client/preference_setup/preference_setup.dm
@@ -287,6 +287,7 @@ var/const/CHARACTER_PREFERENCE_INPUT_TITLE = "Character Preference"
 				pref.client = C
 				break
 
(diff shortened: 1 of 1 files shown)


Jamini (approved):
One small non-modular change. If It fixes the turbo-lag during chargen I'm all for it.

If you break my character generation screen on linux I'm gonna break your kneecaps though, catboy.

PR #78, opened 2020-12-14
Smol drone change

Ikmachek:
## About The Pull Request
Just changes the flavor text on repair drones.

## Why It's Good For The Game
Because NT doesn't literally make everything ever in this setting.

## Changelog
```changelog
tweak: Tiny drone flavor text fix
```

### DIFF
--- a/zzzz_modular_occulus/code/game/occulus-lore-fixes.dm
+++ b/zzzz_modular_occulus/code/game/occulus-lore-fixes.dm
@@ -184,3 +184,7 @@
 /mob/living/carbon/superior_animal/roach
 	name = "Kampfer Roach"
 	desc = "A monstrous, dog-sized cockroach. These huge mutants can be found everywhere on the ship."
+	
+//Drones
+/mob/living/silicon/robot/drone/init()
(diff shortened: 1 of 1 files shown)


ShadzFen (approved):
Small flavortext addition. Modular. All good

ShadzFen:
Ideally it would be cool to have the drone flavortext depend on the player's input. It could be a fun thing to add when we rework our character customization. Have a dedicated flavortext for drones, so there can be multiple unique repair drones. Maybe even force them to input a custom ID number so we can see the same repair drone every now and then, which would add to immersion.

cebutris:
> Ideally it would be cool to have the drone flavortext depend on the player's input. It could be a fun thing to add when we rework our character customization. Have a dedicated flavortext for drones, so there can be multiple unique repair drones. Maybe even force them to input a custom ID number so we can see the same repair drone every now and then, which would add to immersion.

I... really don't think that's the direction we should be taking maint drones. They're not supposed to be characters. They're little robots that have a basic brain to fix problems. If you want to have a robot character with flavor text, we have borgs for that.

ShadzFen:
Eh, fair enough. Another idea could be to have multiple different flavortexts that get randomly assigned when a new drone spawns. So instead of all of them looking the same, you could spawn with one that's a little more rusty, or one that looks brand new, or even one that's been vandalized with graffiti by one of the vagabonds. I dunno, just something to make them more interesting to look at.

PR #81, opened 2020-12-15
Ports the Revelation litany from upstream

Michiyamenotehifunana:
## About The Pull Request

As we did not get the individual objectives update from upstream, we have missed out on two litanies, Revelation and Sanctify. Revelation causes a hallucination in the target, and has a chance to either restore or drain their sanity. Sanctify is purely fluff and literally does nothing aside from completing individual objectives. This sanctify should NOT be confused with the church's weapon's sanctify upgrade, which does something else entirely.

This PR will only port Revelation, and tweaks it slightly. It will have generally less negative effects, but won't be as powerful.

## Changelog
```changelog Toriate
add: Ported the revelation litany from upstream.
```

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/base.dm
+++ b/code/modules/core_implant/cruciform/rituals/base.dm
@@ -111,3 +111,22 @@
 		fail("No target. Make sure your target is either in front of you or grabbed by you.", H, C)
 		return FALSE
 	return TRUE
+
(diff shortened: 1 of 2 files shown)


ShadzFen (approved):
Small upstream thingy. Non modular as most upstream stuff is. Modified slightly to be better for our server. All good.

PR #83, opened 2020-12-17
Ikky messes with global.dm

Ikmachek:
## About The Pull Request
AI laws were set only to protect NT persons, I did a little bit of digging to find what the var [company name] was, found a underloved datum file and while that wasn't the source of the var, I'm keeping it!

## Why It's Good For The Game
Only 2 departments shouldn't be crew to silicons.

## Changelog
```changelog
tweak: fixes a few global vars for lore compliance.
```


Jamini (changes requested):
Ikky, you need to mark *each* non-modular line with "//Occulus Edit"

OR

Use 

//Occulus Edit Start
//Occulus Edit End

For the start and end of your changes.

Jamini (approved):

PR #94, opened 2020-12-21
Enables Object Depossession

Michiyamenotehifunana:
## About The Pull Request

Non-modular change to enable the Release Obj verb which is what you actually need to use to leave an item after using the Possess Obj Verb.

## Changelog
```changelog Toriate
admin: you can now actually depossess objects rather than becoming permanently stuck in one
```


ShadzFen (approved):
Finally! Thank you for this. All looks fine. Get this in, we really do need it. Non-modular.

PR #95, opened 2020-12-22
Adds a christmas hat and adds it to the loadout

Michiyamenotehifunana:
## About The Pull Request
cnZfDD7xuF
VuzonFZney

Sprites hastily scribbled up in 5 minutes by me, and licensed under CC-BY-SA. (that means if you steal these sprites for whatever, you'll have to credit me, and if you modified the sprite, that modification will also need to be shared under CC-BY-SA and you have to credit me for the original thing [but why would you steal these sprites anyways you can probably make a better one if you had more time to spend])

We're long overdue for christmas content! This is the bare minimum, but it'll hold us off for now.

## Changelog
```changelog Toriate
add: Added a christmas hat to the loadout
```


Jamini (approved):
Modular and just in time for the holidays.

PR #98, opened 2020-12-24
Fixes a few miscellaneous mapping bugs

bearrrrrrrr:
## About The Pull Request

Makes the following fixes:

-  No more erronious double wall next to the newly-added maint area.
-  Moved the newly-added maint area door into a location that actually makes snese.
-  Fixes centering on a wall-mounted O2 tank that, as far as I can tell, has been broken for...literally a year?
- Adds a tile underneath a shutter next to the newly-mapped dorms. It previously only had space underneath it, and as such was improperly lighted and tiled.

## Why It's Good For The Game

Fixes good.
Pics below, in order of list.

## Changelog
```changelog
fix: fixed a few mapping bugs.
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -87303,7 +87303,9 @@
 /turf/simulated/open,
 /area/eris/neotheology/chapel)
 "dWx" = (
-/obj/structure/closet/wall_mounted/emcloset/escape_pods,
+/obj/structure/closet/wall_mounted/emcloset/escape_pods{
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
Moves the door further away from the airlock. Was so far down the hall that it was out of sight til quite a while down. Oops...
This is also a pic of the double wall (was a wall above where the stash was, looked very bad)
image

Improper centering on O2 wall thing, was previously incenter of tile.
image


Improper tile: (top left shutter, put plating undernearth it)
image

Jamini (approved):
Everything looks good!

PR #103, opened 2020-12-25
Nerfs the winebath

Michiyamenotehifunana:
## About The Pull Request

The winebath can no longer heal brute or burn damage below 50. In exchange, it gets some major bloodclotting effects.

Entering the winebath will bloody your clothes with wine.

Remaining in the winebath will make you slur and see rainbows. I have opted against using the dizzy effect as that shakes your screen and can cause motion sickness.

## Why It's Good For The Game

Church got a lotta buffs, so its time they got a nerf.

## Changelog
```changelog Toriate
balance: nerfed the winebath
```

### DIFF
--- a/code/modules/core_implant/cruciform/machinery/vat.dm
+++ b/code/modules/core_implant/cruciform/machinery/vat.dm
@@ -83,6 +83,9 @@
 		O.loc = loc
 	add_fingerprint(user)
 	buckle_mob(C)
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Nothing unmentioned. Everything is marked where it is non-modular. Should be nice to see in-game without it feeling too unfair.

PR #107, opened 2020-12-29
Fixed pickweight() selecting options with a weight of 0.

Jamini:
## About The Pull Request

Weighted list sorting will now no longer select options that have 0 weight.

## Why It's Good For The Game

A weight of 0 means something should never be picked.

## Changelog
```changelog
fix: adjusted pickweight() proc to be more sane
```


Michiyamenotehifunana (approved):
Looks basic enough, all good!

PR #108, opened 2020-12-31
Fixes a runtime and makes oddities have perks again!

Michiyamenotehifunana:
## About The Pull Request
There's a PR upstream that straight out nerfs oddities and makes it so that they only have a limited chance to spawn with perks, which was seemingly in place already when the EOTP stuff was done upstream, hence us getting some changes that unintentionally broke the crap out of our oddity perks. Until we decide to port that, we'll have to revert to the old system.

This will also fix a runtime with surgery'ing detached limbs.

## Changelog
```changelog Toriate
fix: Oddities should have perks again!
fix: Surgery'ing detached limbs should not throw an error anymore!
```


Jamini:
Everything seems in place.

Jamini (approved):

PR #112, opened 2020-12-31
Maps in an underwear drawer, deletes extraneous bar chairs

Michiyamenotehifunana:
## About The Pull Request

Exactly what it says on the tin. No more chairs on top of sofas, and the showers now gets an underwear drawer.

## Changelog
```changelog Toriate
add: An underwear drawer has been installed in the showers!
fix: The club should no longer have chairs installed on top of the sofas. The contractors responsible for that particularly innovative solution have been sacked.
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -49226,9 +49226,6 @@
 /turf/simulated/floor/tiled/steel/bar_dance,
 /area/eris/crew_quarters/bar)
 "ckL" = (
-/obj/structure/bed/chair/custom/bar_special{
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Looks good.

ShadzFen:
Map changes would be appreciated with pictures to show the changes.

Jamini:
> Map changes would be appreciated with pictures to show the changes.

Note really needed in this case since it's a very small PR. You can see which tiles are getting edited.

ShadzFen:
Yes but not everyone can read code. We're not the only ones who check these PRs! When it comes to DMI or map changes, pictures should be standard.

PR #115, opened 2021-01-01
Fixed nonhumans getting doubled organs on clone

Jamini:
## About The Pull Request

Fixes a bad proc call.
Tested and confirms work
High priority fix

## Why It's Good For The Game

High prioirty fix
Prior to erismed2 you couldn't have duplicate organs, so it wasn't neccesary to delete something unless we wanted to replace it (in character_prefs). Now that we can have duplicates, we need to delete the organs created by New() before adding in all the organs a species should have!

TL:DR- We are kinda dumb and this is an issue based on our code!

## Changelog
```changelog
fix: cloning no longer doubles all your organs.
```


ShadzFen (approved):
Looks good. non-modular bugfix

PR #123, opened 2021-01-05
Add 10 slots as requested

Jamini:
## About The Pull Request

Ten more character slots for people to fill with faces they never play.

## Why It's Good For The Game

A simple request

## Changelog
```changelog
add: ten character slots
```


Jamini:
Nevermind. Not needed. This is a config option

PR #127, opened 2021-01-09
Soulcrypt Antispam Tweaks

Michiyamenotehifunana:
## About The Pull Request
Now the soulcrypt should spam you less! Also added an indicator that your SC got EMP'd.

## Changelog
```changelog
tweak: SCs should now spam you less. They should also notify you if it has been EMP'd now.
```


Jamini (approved):
Mostly tori updating SC stuff... it should probably be marked since it's nonmodular though

PR #134, opened 2021-01-09
Events with a weighting of 0 will no longer fire.

Jamini:
## About The Pull Request

Events with a weighting of 0 will no longer fire

## Why It's Good For The Game

0 weighting means SOMETHING SHOULD NOT BE PICKED

## Changelog
```changelog
fix: Major events such as Hivemind should no longer trigger during a Healer round.
```


ShadzFen (approved):
Looks good enough. If it's set to 0 then it probably isn't meant to spawn to begin with. Non-modular small code addition, good to go.

PR #135, opened 2021-01-10
Re-rebalances the EC9

Michiyamenotehifunana:
## About The Pull Request

This makes the EC9 take SMG magazines again, increases its unwielded recoil penalty a bit, and nearly doubles its spawn rarity.

## Why It's Good For The Game

So it turned out I had somehow made it 2x as common as the Atreides and then kept thinking I had made it the same rarity as the Atreides. It is now only slightly more common than the Atreides. Gone are the days of EC9s dominating maint loot!

It has also gotten the ability to use SMG magazines back after some more player feedback, and its recoil has been boosted a bit to be more inline with all the other SMGs.

This should make the EC9 less of a sore thumb in the game overall.

## Changelog
```changelog Toriate
balance: rebalanced the EC9 so that it will stop dominating maint loot and be more inline with all the other SMGs in the sense that it actually takes SMG magazines now.
```


Jamini (approved):
This one is your baby Tori. looks good from this end

PR #136, opened 2021-01-10
Unincludes bear's new dungeon from the main .dme

Michiyamenotehifunana:
## About The Pull Request
Exactly what it says on the tin.

## Why It's Good For The Game
Deepmaint rooms do not need to be compiled with everything else, as a different mechanism is responsible for loading them when deepmaint is generated.

At best, this would slow down regular compiling ever so slightly. At worst, this would break it somehow (unconfirmed but you'll never know with BYOND.).

## Changelog
```changelog Toriate
code: bear's new deepmaint room is no longer included
```


Jamini (approved):
Keeping stuff clean is good.

bearrrrrrrr:
$ git blame-someone-else

PR #139, opened 2021-01-13
Basis for hooded clothes

RadiantFlash7:
## About The Pull Request
Adds the framework for any further hooded clothing. 
Also adds my totally snowflake hooded jacket to the loadout. (Should be sprite consistent)

## Why It's Good For The Game
Hooded clothing! Style points, and atmosphere. Could also add hooded church clothing in the future, which would be a pretty cool flavor for Mehkane.

## Changelog
```changelog
add: Added hooded jacket code, and the sleeveless jacket item into loadout.
```


zzzz_modular_occulus/code/modules/client/preference_setup/loadout/lists/suits.dm
```diff
@@ -0,0 +1,11 @@
+// Suit slot
+/datum/gear/suit
```
Michiyamenotehifunana, line 2: This bit doesn't need to be copypasta'd! You can comment it out if you wanna keep it for future reference without looking for the original file.

zzzz_modular_occulus/code/modules/clothing/suit/hooded.dm
```diff
@@ -0,0 +1,78 @@
...
+				hood.canremove = FALSE
+				icon_state = "[toggleicon]_t"
+				H.update_inv_wear_suit()
+	else
+		RemoveHood()
+
+/* For anyone adding hooded clothes, take note.
+You need 2 versions of the sprite, one with the normal icon and one with icon name and '_t' at the end of it,
+for both Item and mob sprited versions.   EX: sleeveless_red & sleeveless_red_t
+Else, it won't work properly.**//
```
Michiyamenotehifunana, line 65: You added one extra * and /!

Michiyamenotehifunana (changes requested):
Some quick tweaks here and there and it should work out fine!

zzzz_modular_occulus/code/modules/clothing/suit/hooded.dm
RadiantFlash7: Fixed this!

zzzz_modular_occulus/code/modules/client/preference_setup/loadout/lists/suits.dm
RadiantFlash7: ...I got entirely confused on how the load out system works, and I copy pasted it without thinking. Whoops.

Michiyamenotehifunana (approved):
Looks good!

Michiyamenotehifunana:
Aseprite_xq5put2hRq 
Sprites for reference

RadiantFlash7:
For reference, The hoods will block ears and anything extending past the head while the hood is up.

PR #141, opened 2021-01-14
Bucket Transfer Change

Leon9621:
Added 50 and 100 to the bucket transfer values, because my OCD was killing me.

## About The Pull Request

Added reagent_containers/glass/beaker.dm to the modular Occulus folder
Added 50 and 100 unit transfer values to the common plastic buckets.

## Why It's Good For The Game

Personally, not having an even amount of water to pour into a tray to refill it bothered me. Previously, your best bet was to use 20, which involved 5 clicks to fill a tray and ten to refill it, or 60 if you didn't mind wasting some water. Not that it's exactly possible to waste what is effectively an infinite resource, but it bothered me enough that I decided it would be a good idea to change it in my first PR. 

Tested it out after, every other transfer value showed up, and the new values worked just fine for taking water in and pouring it out.

## Changelog
```changelog
tweak: Buckets can now transfer 50 / 100 units at a time, in addition to their previous values.
```

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -2836,6 +2836,7 @@
 #include "zzzz_modular_occulus\code\modules\projectiles\projectile\bullettypes.dm"
 #include "zzzz_modular_occulus\code\modules\reagents\food-Drinks.dm"
 #include "zzzz_modular_occulus\code\modules\reagents\recipes.dm"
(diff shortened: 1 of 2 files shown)


Michiyamenotehifunana (approved):
Very simple changes, looks good!

PR #142, opened 2021-01-15
Revamps/remaps the space fortress into a One Star location.

bearrrrrrrr:
## About The Pull Request

Revamps/remaps the space fortress into a One Star location. Also changes some of the lines on the Type 34C, as well as setting needs power = 0 to the abandoned fortress. Probably will be adding or removing things depending on how hard it is. I don't want to make it too easy, now.

## Why It's Good For The Game

Adds an area with one-star tech and tools, stuff that I don't think we've ever gotten. You'll still have to fight through what is basically the fucking gauntlet in terms of mobs, though, so...good luck.

## Changelog
```changelog
add: Remaps the abandoned fortress
```


bearrrrrrrr:
image
Screenie, although you can't really tell much from it.

bearrrrrrrr:
good for review proper

Jamini:
Seems fun. One concern: What are the green tiles?

bearrrrrrrr:
> Seems fun. One concern: What are the green tiles?

Grass, in-line with other OS areas that include nature-like decorations.

code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
```diff
@@ -117,7 +117,7 @@
 	for(var/mob/M in range(10,src))
 		if(M.client)
 			shake_camera(M, 4, 3)
-	visible_message(SPAN_DANGER(pick("Prepare to die!", "JUSTICE", "Run!")))
+	visible_message(SPAN_DANGER(pick("JUDGING...", "THE HAND OF FATE DECIDES", "THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES THE SKIES ")))
```
ShadzFen, line 120: I'm worried "THE SKIES" might spam the chat a little too much. Is one of these only spoken once or multiple times?

bearrrrrrrr: These are messages that the big bad boy says to choreograph when he's about to use a big bad attack, hence why the length on all of them are increased. It also doesn't spam the chat as much as you'd suspect! Can still reduce it if you want, but I don't think I need to.

Jamini (changes requested):
your change in code/game/area/asteroid_areas.dm needs to be marked.
As does the megafauna lines line

bearrrrrrrr:
Sorry! Fixed. Brain fart.

Jamini (approved):

PR #153, opened 2021-01-18
Makes rubbershot actually rubber

Michiyamenotehifunana:
## About The Pull Request

Rubbershot is no longer sharp and will not embed, and will also not penetrate armor.

## Why It's Good For The Game

Rubbershot is supposed to be non-lethal. Embedding is lethal. Causing bleeding is lethal.

## Changelog
```changelog Toriate
fix: rubbershot is now actually rubber
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/projectiles/projectile/bullettypes.dm
+++ b/zzzz_modular_occulus/code/modules/projectiles/projectile/bullettypes.dm
@@ -9,6 +9,9 @@
 	range_step = 2
 	spread_step = 10
 	knockback = 1
+	armor_penetration = 0
+	embed = FALSE
+	sharp = FALSE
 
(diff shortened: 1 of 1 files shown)


Leon9621:
Rubber rounds are still absolutely capable of breaking skin and causing permanent damage irl. Which is a shitty argument to make on it's own, but it's not like we actually lack any nonlethal options (tasers, tear gas, stun gloves / stunbatons, etc). 

To get my two cents in here all proper like, I'd like to see what we could do to reduce the chance of the rubbershot embedding, as opposed to removing it outright.

bearrrrrrrr:
tbh if you remove the embedding aspect you probably want to majorly increase the damage, just imo?
iunno

Jamini (approved):
Emojicracy approved. We are also likely to remove the knockback on rubbershot as well.

PR #155, opened 2021-01-18
Fix RCON input load display

### LINKED ISSUE #137: RCON fails to show power input.
#### Description of issue

image

These SMESs are charging and receiving power. However, unless you go and see them in person, the RCON displays 0 input charge.

It should show the amount of charge the SMES is receiving.

#### Steps to reproduce 

Start engine. Check Engine Room SMESs (Partially or fully charging). Check RCON (0kw Input)

At least a week.

Server revision: master - 2021-01-10
a5dcecc59d76ab0225e5e0240cc046b881d47dd4

#### Issue bingo

zackerycodestoo:
## About The Pull Request

Fixes #137

This is a small edit to the RCON console code to fix a bug where the Input Load line was always at zero. The cause of the bug is that the RCON console was attempting to pull from an SMES variable called input_available, which was not actually defined beyond always being set at zero. I've instead pointed the RCON console to look at a variable called target_load, which outputs exactly how much the SMES is trying to draw from the network at a time (this does zero out/show smaller numbers when the SMES is full/nearly full.)

## Why It's Good For The Game

Engineers will now be able to pinpoint whether power issues are sourced from a substation not receiving power rather than manually checking the wires to the substation.

## Changelog
```changelog
fix: RCON console now properly shows input load (previously always read zero)
```

### DIFF
--- a/code/modules/modular_computers/file_system/programs/engineering/rcon_console.dm
+++ b/code/modules/modular_computers/file_system/programs/engineering/rcon_console.dm
@@ -33,7 +33,7 @@
 		"charge" = round(SMES.Percentage()),
 		"input_set" = SMES.input_attempt,
 		"input_val" = round(SMES.input_level/1000, 0.1),
-		"input_load" = round(SMES.input_available/1000, 0.1),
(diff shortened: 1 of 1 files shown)


ShadzFen (approved):
Small non-modular code fix. Marked. All good

PR #156, opened 2021-01-19
Adds a tcomms relay to the science shuttle

Michiyamenotehifunana:
## About The Pull Request

Exactly what it says on the tin.

## Why It's Good For The Game

More often than not, the science shuttle gets used for rescuing dead miners, so being able to coordinate with everyone else should make things a lot less hectic. It should also reduce incidences of the ship leaving the shuttles behind, although that has never happened before as far as I'm aware.

## Changelog
```changelog Toriate
add: The Vasily Dorkuchev has been retrofitted with a state of the art telecommunications relay to facilitate communications with the Northern Light.
```


Jamini (approved):
This makes sense and probably should have been done before.

PR #157, opened 2021-01-19
Halves the nutrition cost of most litanies, and fixes some bugs

Michiyamenotehifunana:
## About The Pull Request

Exactly what it says on the tin. A bug has also been fixed where the Scrying litany logs the user as looking through their own eyes when that is in fact impossible and they're looking through someone else's eyes. Also, the toxloss of Soul Hunger has been reduced juuust a bit.

## Why It's Good For The Game

Emojicracy voted in favor of halving all litany nutrition costs. Should make them far more useable.

## Changelog
```changelog Toriate
balance: Litany nutrition costs have been toned back!
```


Jamini (approved):
Emojicracy approved reductions to the lit nerfs we did.

PR #158, opened 2021-01-19
Adds a .40 lever action rifle

Leon9621:
Adds a lever action rifle to the game, with a sawed off variant and a functional design disk that isn't currently obtainable.

## About The Pull Request
Adds a lever action rifle that fires .40 magnum to the game, with a sawn off variant. The rifle is made of steel and plastic (18 and 6 respectively) and sells for cheap due to the lack of any expensive materials used to create it.

The disk is pretty standard, print the gun for 3 points, get boxes of ammo for 1 per.

## Why It's Good For The Game
Lever actions are cool! Plus it's a hopefully decent budget gun if plasteel is lacking for any reason.

## Changelog
```changelog
add: Added new things
```


Leon9621:
And I just realized that I commented stuff out instead of deleting it, for the things it inherits from the lever action shotgun. I can fix that when I get home.

Jamini (dismissed):
Seems strictly weaker than a lot of stuff, which is good. Should export for 180cr (600*0.60/2) which is about in-line with circular saws.

Leon9621:
Yeah, I definitely don't want to drop anything too ridiculously strong or anything in here. It doesn't have the nutty damage / penetration increases that the handguns have, but it can fire relatively quickly and, of course, you can put different mods on it.

Jamini (approved):
Stillfine

PR #159, opened 2021-01-20
Fixes custom named item losing name when wielded

ShadzFen:
## About The Pull Request

Custom named items will now keep their custom name after wielding and unwielding them. Small little bug fix that was missed during testing and was more of an overlook from the original code than anything.

image

## Why It's Good For The Game

People's custom named items will no longer be set back to the original item's base name, and instead will keep their custom name!

## Changelog
```changelog
fix: Custom Items will retain their custom name after unwielding them.
```

### DIFF
--- a/code/game/objects/items/wielding.dm
+++ b/code/game/objects/items/wielding.dm
@@ -23,9 +23,9 @@
 		force = force_unwielded
 	else
 		force = (force / 1.3)
-	var/sf = findtext(name,"Wielded ")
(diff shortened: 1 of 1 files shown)


Jamini (changes requested):
Shaaaaadz, you forgot to MARK YOUR NON-MODULAR CHANGES!

ShadzFen:
Heck, I forgot, hold on. I haven't actually written code in several months... Yes I'm putting excuses

Jamini (approved):

PR #163, opened 2021-01-20
Tunes up the Lever Action, clarifies it's caliber.

Leon9621:
## About The Pull Request
Clarifies it's ammo and makes it a _little_ more competitive. I don't want to kneejerk this thing into being way too overtuned.

## Why It's Good For The Game
I worried about this basically being a worse revolver, and (the only) feedback I've gotten so far kinda confirmed that. Every other revolver has speedloaders, very nice damage / armor pen multipliers, and fits in holsters, and this was... not better than them in any way at all. Hopefully this helps out a little.

## Changelog
```changelog
balance: rebalanced the lever action rifle
```


Michiyamenotehifunana (approved):
Looks great for a start!

PR #164, opened 2021-01-22
Ports slot machines

Michiyamenotehifunana:
## About The Pull Request
Ports https://github.com/discordia-space/CEV-Eris/pull/5319 and maps it in.
StrongDMM_70oCXNZUIG
StrongDMM_DLkxRIMBYu
dreamseeker_nefSBMf2Pc
dreamseeker_yDXeDF5MmH

## Why It's Good For The Game

More fun things to waste money on! Gamble!

## Changelog
```changelog Toriate
add: Ports slotmachines from upstream.
```


ShadzFen:
Awesome! Though we might want to discuss the location at some point, with Jams too, see what he thinks. I feel like it would fit better downstairs rather than right next to an ATM. It also turns that upper section hallway into a single tile hallway, which is never good.

Michiyamenotehifunana:
We should get Ash back in the discord first so we can properly plan out how to jam this thing into the lower floors, since they spearheaded the redesign of the bar to begin with.

Jamini:
Where does the revenue from the slot machine go? To which account?

Jamini:
I'm also really not a fan of making the top floor even more crowded. Could it go in the smoke area on the lower level?

Michiyamenotehifunana:
There is no revenue from the slot machine. It is simply added to its jackpot, and wins are paid out from its jackpot.

I'd also rather get someone else who's more passionately involved with the club's layout to move it to an appropriate place, perhaps after this PR is merged.

Jamini (approved):
Fair enough. Everything looks good from this end.

PR #168, opened 2021-01-23
Adds and maps in the advanced roller bed!

Michiyamenotehifunana:
## About The Pull Request

Aseprite_fiLhgaoA8t

A roller bed that can finally fit in your bag when collapsed! Basically SMG sized. Also fits in medical belts.

Only two are available in the paramedic's closet, with no way to get more.

StrongDMM_2Q12HGu6ec

## Why It's Good For The Game

Players have been asking for rollerbeds that can fit in bags since forever! I figured I'd instead add an advanced variant that's scarce instead of making all the rollerbeds fit.

## Changelog
```changelog
add: The NanoTrasen Procurement department has recently rediscovered a cache of ultra-compact folding beds in a warehouse formerly owned by a competitor which was bought out an indeterminate amount of time ago. The paramedics closet has been supplied with two specimens.
```


ShadzFen:
Is it not bad to have an item that can't be made, even with high level R&D? Genuine question, I dunno. I do like the idea of them being that level of scarce, like that they might get used more carefully.

Michiyamenotehifunana:
It gives them near relic-levels of rarity, so it *should* balance it out if its small size is too overwhelmingly powerful.

ShadzFen (approved):
New modular addition. All good! Lets hope the small size is compensated by the uniqueness of there only being 2 on the entire ship with no possibility to manufacture or buy additional ones.

Jamini:
I'll merge this, but we already approved making this an RnD item. So you can dd it to one of the biotech nodes @Toriate

PR #169, opened 2021-01-23
Alt Civ Pack Balance Pass

Michiyamenotehifunana:
## About The Pull Request

The EC9 and RX8 have been toned down quite a bit so that they are not strictly superior to the Atreides. In particular, both of them have had their damages slashed by a rather large margin. I had been balancing them against semi-auto weapons so their numbers were out of whack. Now with reference to other full auto SMG class weapons, they should be less silly.

## Why It's Good For The Game

Balance is good!

## Changelog
```changelog Toriate
balance: rebalanced the EC9 and RX8 to be more in line with SMG-class guns. A chunk of their damage has been taken away.
```


ShadzFen (approved):
Some modular variable changes. All good

PR #173, opened 2021-01-25
Makes body markings not lag anymore

Michiyamenotehifunana:
## About The Pull Request

By shifting body marking rendering onto limb icon generation code, lag from excessive markings seems to have been drastically reduced, at least on my local server. Previously, body markings were tacked on to the overall mob icon after everything was rendered which seemed to really screw with the icon cache system.

Of course, there is still a chance that the server will still chug, but it should hopefully be reduced rather than made worse with these changes.

## Why It's Good For The Game

Lag reduction is good. Body markings freezing and lagging the shit out of the game has been a massive problem since forever.

## Changelog
```changelog Toriate
refactor: Body markings have been moved over to limb icon generation code, which should reduce the amount of lag caused by body markings.
```

### DIFF
--- a/code/game/dna/dna2_helpers.dm
+++ b/code/game/dna/dna2_helpers.dm
@@ -173,6 +173,14 @@
 		//Body markings
 		H.body_markings = dna.body_markings.Copy()
 
+		// OCCULUS EDIT IN AN ECLIPSE EDIT - HONK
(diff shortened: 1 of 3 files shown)


Jamini (approved):
I see nothing objectionable. Some images of this functioning in-game would be nice, but not required.

PR #175, opened 2021-01-25
Fix mech drills and improve feedback

### LINKED ISSUE #171: mining Exosuits are useless due to mining exo suit drill bits breaking
#### Description of issue
Exosuit drill bits are useless and break in one use.

Steel bits, Even plasteel bits. All break on first use. Makes mining exosuits /completely/ useless.

#### Steps to reproduce 
mine with an exosuit, watch as your drill bit breaks on first use. Even expensive fancy drillbits. Cry inside.

The plasteel drill head breaks apart!

As long as I've been on occulus

Server revision: master - 2021-01-24

#### Issue bingo

- [ ] [Couldn't find an existing issue about this](https://github.com/discordia-space/CEV-Eris/issues)
*Edit: Fixed the boxes for you.

zackerycodestoo:
## About The Pull Request

Fixes #171

Currently, mech drills are completely useless because they shatter after the first digging. This is not intended behaviour!

The issue with the drills is that when their durability is determined, it's calculated at a time when the material hasn't yet been set -- so instead of the proper durability, the drill is given a durability of 1.

A verb was built in (it may be able to be put in the modular folder but I couldn't figure it out) to do ApplyDurability() to make the drill head get the proper durability. Newly spawned mechs also receive the proper drill durability.

A final small change is to add a sound when the drill starts drilling and when it finished for more satisfying player feedback. 

image
image

## Why It's Good For The Game

Miners can use mech drills finally! A stronger drillbit drills faster and more powerfully. A single drillbit may be able to last a whole mining mission, but plasteel and diamonds can be made in the field by miners for even more efficiency.

## Changelog
```changelog
tweak: Mech drills have more sound feedback.
fix: Mech drills receive durability properly now and can be used for extended mining missions.
```

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -204,13 +204,24 @@
 
 /obj/item/weapon/material/drill_head/Initialize()
 	. = ..()
-	durability = 2 * (material ? material.integrity : 1)
+	// OCCULUS EDIT -- durability was not being properly applied because material during this proc is null
+	//durability = 2 * (material ? material.integrity : 1)
+
+/obj/item/weapon/material/drill_head/steel/New(var/newloc)
+	..(newloc,MATERIAL_STEEL)
+	src.ApplyDurability()	// OCCULUS EDIT -- apply durability to drills properly
 
 /obj/item/weapon/material/drill_head/plasteel/New(var/newloc)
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana (approved):
Everything looks about right!

PR #177, opened 2021-01-26
Fix minerals so they can be seen through walls with mesons

### LINKED ISSUE #37: Meson Scanners do not allow viewing ore veins as intended
#### Description of issue

Meson scanners are not able to see ore through rock walls

Original intention of Meson scanners was to see ore deposits through walls.
Eris changed this at some point.
We need this fixed/reverted to intended behavior.

#### Steps to reproduce 

1. Equip meson scanners
2. Go to the asteroid
3. Activate mesons
4. See deposits through walls.

This is likely caused because ores are now an overlay rather than a type of wall. We may need to implement some spaghetti code or modify mesons so they can see overlays. Whichever is cleaner.

Uncertain, but likely since ore was reworked years ago

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [ ] [Couldn't find an existing issue about this](https://github.com/discordia-space/CEV-Eris/issues)

zackerycodestoo:
## About The Pull Request

Fixes #37

Thanks to a big tip from RadiantFlash, I was able to narrow down where the issue was. During generation, minerals don't add themselves to the overlays like archaeological spots do -- now, minerals can be seen through walls while using mesons.

image
image

## Why It's Good For The Game

Miners can see where to dig now!

## Changelog
```changelog
fix: Mesons now see ore through walls
```

### DIFF
--- a/code/modules/mining/mineral_effect.dm
+++ b/code/modules/mining/mineral_effect.dm
@@ -15,7 +15,7 @@
 	ore_key = M.name
 	icon_state = "rock_[ore_key]"
 	var/turf/T = get_turf(src)
-	T.overlays += image('icons/obj/mining.dmi', "rock_[ore_key]", dir = 1)
+	T.overlays += image('icons/obj/mining.dmi', "rock_[ore_key]", dir = 1)	// OCCULUS EDIT: Show ores through walls when using mesons
 	if(T.color)
 		color = T.color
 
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Nice catch Mr.Fox.

PR #181, opened 2021-01-28
Revert "Makes body markings not lag anymore"

Jamini:
Reverts Occulus-Server/Occulus-Eris#173


Michiyamenotehifunana (approved):
We're all out of ideas so lets ram this through

PR #182, opened 2021-01-28
Revert "Stop meteors from spawning pre-round and Runlevel fixes"

Jamini:
Reverts Occulus-Server/Occulus-Eris#166


Michiyamenotehifunana (approved):
Emergency approval on the basis that we have no idea what we're doing and we're panicking

PR #185, opened 2021-01-28
Revert "Revert "Makes body markings not lag anymore""

Michiyamenotehifunana:
Reverts Occulus-Server/Occulus-Eris#181


Jamini (approved):
Apologies for not doing this myself. Approved and speedmerged. It should be in tonight.

PR #186, opened 2021-01-28
Stop pre-round meteors (without runlevel changes)

zackerycodestoo:
## About The Pull Request

This is a resubmission of #166 without fixes to the Runlevel in the Master Controller.

The code will instead look specifically for Runlevel 3 (the round has started) before determining if meteors can spawn. This is somewhat misleading, since the Master Controller reports that 'the round has started' is actually Runlevel 4, which the Master Controller variables themselves show that it is Runlevel 3.

The previous attempt to fix what the Master Controller believed the Runlevel is ran into unforeseen consequences where the Storyteller system did not tick properly, among other issues.

Testing log:

1. Started game, watched Storyteller Panel to make sure event pool points were being gained. Points were being gained successfully.
2. During runlevel 1/2 (setup/pre-game lobby): waited for meteor to attempt to spawn, successfully received failure message.

image

3. During runlevel 3 (round has started): force spawned meteor shower via Storyteller Panel, admin log had impacts as expected.

image

Finally, the actual insertion of the conditional clause to stop the meteors was moved to a more precise point in the event that a failure to return might cause further problems.

## Why It's Good For The Game

There are times when the server sits with no players on it! During this time, meteors can still spawn and damage the ship before the game begins. If someone wants to play and they don't want to repair all the damage, this usually leads to a transfer vote to start a fresh ship.

This change does not prevent any meteors from spawning immediately after the game starts -- only before it starts.

## Changelog
```changelog
fix: Meteors no longer impact the ship before the round begins.
```

### DIFF
--- a/code/game/gamemodes/events/meteors.dm
+++ b/code/game/gamemodes/events/meteors.dm
@@ -190,11 +190,12 @@
 	var/Me = pickweight(meteortypes)
 	var/obj/effect/meteor/M = new Me(pickedstart)
 	M.dest = pickedgoal
-	if (Master.current_runlevel >= 3)	// OCCULUS EDIT: 3 is the actual number for RUNLEVEL_GAME -- adjusting the Runlevel code
-										// to change it to 4 was met with unforeseen consequences
+	///// OCCULUS EDIT BEGIN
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana (approved):
Looks good! Non-modular but marked!

PR #192, opened 2021-01-29
Fix crafted drill heads

### LINKED ISSUE #188: Mech drill replacement bits have incorrect durability
#### Description of issue

Crafted drill bits are not spawning with the correct durability. I am uncertain why this is happening since I confirmed they had the correct durability during my testing. This issue is primarily a reminder for myself to look at the code again in the morning.

Drill heads produced from mech drills work with the proper durability as expected.

Steel drill heads should spawn with 200 durability, but they spawn with 0.

#### Steps to reproduce 

1. Craft a steel drill head from the crafting menu
2. Durability is 0

Utility.dm starting at 198. ApplyDurability() proc is seemingly not working as expected (proc defined at line 223)

Presumably ever since new drill durability fix was implemented on Wednesday Jan 27

image

#### Issue bingo

- [X] Issue could be reproduced by different players

zackerycodestoo:
## About The Pull Request

Fixed #188

This is a slight correction to #175. In previous testing, I spawned drill heads directly and spawned mech drills and they had the correct durability. I operated under the assumption that making drills from the crafting menu would also work since spawns did, but I did not test thoroughly enough.

Crafted drills call Created() rather than another proc. This fix makes it so drill heads crafted through the crafting menu have the proper durability.

image
image

Crafting of plasteel and diamond drill heads were also tested and received durability as expected.

## Why It's Good For The Game

Miners can now make field replacements for drill heads that actually work.

## Changelog
```changelog
fix: Drill heads crafted through the crafting menu now get the proper durability.
```

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -207,6 +207,15 @@
 	// OCCULUS EDIT -- durability was not being properly applied because material during this proc is null
 	//durability = 2 * (material ? material.integrity : 1)
 
+///// OCCULUS EDIT BEGIN
+// Drills crafted from the crafting menu actually call Created().
+// var/creator is a dummy var to avoid any issues with the call.
+
+/obj/item/weapon/material/drill_head/Created(var/creator)
+	src.ApplyDurability()
+
+///// OCCULUS EDIT END
+
 /obj/item/weapon/material/drill_head/steel/New(var/newloc)
 	..(newloc,MATERIAL_STEEL)
 	src.ApplyDurability()	// OCCULUS EDIT -- apply durability to drills properly


code/modules/mechs/equipment/utility.dm
```diff
@@ -207,6 +207,15 @@
 	// OCCULUS EDIT -- durability was not being properly applied because material during this proc is null
 	//durability = 2 * (material ? material.integrity : 1)
 
+///// OCCULUS EDIT BEGIN
+// Drills crafted from the crafting menu actually call Created().
+// var/creator is a dummy var to avoid any issues with the call.
+
+/obj/item/weapon/material/drill_head/Created(var/creator)
```
Michiyamenotehifunana, line 214: You don't actually need var/creator here! It isn't used by anything else in the code so it'd have no effect here.

Michiyamenotehifunana (approved):
Non-modular but its marked and looks about right!

PR #203, opened 2021-02-03
Changes noble description to add a clear glass clear descriptor on what it does mechanically as a downside besides sanity drain

IrkallaEpsilon:
Yeah no, I dont like bait. It baits with interesting ransoming objectives and all it does in reality is make you TC batteries via the most uninspiring objective ever conceived. This doesnt fix that it just adds a disclaimer/additional description on what this trait actually does.

## About The Pull Request

Changes nobles trait description when you pick it in the character creator. Now it shouldn't hide the fact that it turns you into overglorified non-rechargeable TC batteries.

## Why It's Good For The Game

Transparancy. One shouldn't need to code dive to see if a description is actually true to what it says or blatant false advertising like it was in this case.

## Changelog
```changelog
fix: Noble description enhanced to no longer falsely advertise ransom/abduction objectives.
```

### DIFF
--- a/code/datums/setup_option/backgrounds/fate.dm
+++ b/code/datums/setup_option/backgrounds/fate.dm
@@ -64,7 +64,9 @@
 	desc = "You are of a prestigious family or lineage. Maybe you're self-made. Maybe you're the heir to a fortune.\
 			Either way, you, like everyone else, are stuck aboard the NEV Northern Light. This is not the luxury you are used to. \
 			You're far more likely to be targeted for abduction by... unsavory types. \
-			Stay safe, star-child."
+			Stay safe, star-child. \
+			((Clarification: By abduction the game strictly and exclusively refers to ASSASSINATE objectives, \
+			not the serbian/merc kidnap objectives))"
 
 	perks = list(PERK_NOBLE)
 


bearrrrrrrr:
irkalla i love you but this seems like a salt-PR. the text doesn't really seem thematically in-line with anything else we have and you could have easily just changed the text itself. this is, like...classic salt.

Jamini (changes requested):
Please append "//Occulus Edit" to any non-modular edits.
This can be modularized.
Also I would recommend putting "Abduction or assassination" in the description line above. There is no need for ooc commentary in the perk description.

IrkallaEpsilon:
Closing, reason:

- A toggle that clearly says "Be direct target of assassination/lace theft/permanent round removal? Yes/No" in the character creator would do the same but more efficient without turning a flavorful perk that just gives a roundstart oddity with stats that can be found after a short maint dive while giving a rather considerable sanity drain. 
- Said toggle would mechanically inhibit people from rolling assassinate no matter what if set to "No.". The proposed change to description does not stop the mechanical possibility of rolling target. It also does not stop statistics, meaning if you can roll target somebody at some point will pick it up regardless if you do not wish to be involved in it due to whatever **valid** reasons you might have.
-> Recent vote results showed theres a fraction of people that are heavily in favor of mandatory permanent round removal objectives. Another fraction considers these objectives very disruptive, unfun and to some degree a  rather consierable waste of time for the receiving end.  Thus this simple description change would inhibit nothing to be precise. Citation/Source can be acquired by looking at the crypt removal objective vote at the most extreme option to the right on the reactions with the pirate flag for a symbol and to the left with the big large X symbol.
->> Therefor this wont fix the issue at hand that people are intrinsically unreliable and have an intrinsic desire to upset one another, a clear disclaimer wont fix a clearly mechanical problem. A toggle that makes you not even appear on said bounty lists for that particular objective no matter your trait will to some degree. That way people would be forced to go the extra mile so to say to bother somebody with this particular objective who wish to not be involved in that. Putting in more effort would be a better deterrent for this. 
->>> A compromise solution, that being a toggle, would satisfy everyone overall less instead of one group more and would still not force anyone into things they personally do not enjoy whatsoever.  Coding this however is beyond my scope and too much of an time investment as it currently stands due to several heavy, stressful and time consuming IRL constrains.

Alternatives to consider:

Buff noble significantly to offset the fact that you are overglorified TC batteries and leave the objective tied to noble.

PR #214, opened 2021-02-05
Add range check for mech clamp/drill

### LINKED ISSUE #206: Mech drills work at range and through solid objects.
#### Description of issue
While piloting a mech with a drill equipped, you can click on anything up to 3 tiles away from your mech in the direction it's facing, and the mech will attempt to dig that tile as though it were directly in front of you. If asteroid rock is clicked it will dig out a 3x3 space centered around where you click; if it's a wall, the mech will attempt to drill down the wall. This works through asteroid rock (though I did not test if it worked through other solid objects) and works whether or not you can see the tile you are clicking on (tried with and without mesons and both worked).

Drills should only work on things that are directly adjacent to the mech. Digging out an AoE is fine, but the center of that area should be directly adjacent to the mech, and not 3 tiles away, and certainly not 3 tiles away in an area you can't even see.

#### Steps to reproduce 
1. Get a mech with a drill attachment
2. Face where you want to drill from 2-3 tiles away
3. Click what you want to drill
4. Watch it get drilled, regardless of obstructions, line of sight, or sanity.

I tested this only with the one mech and one drill available to me (/obj/item/mecha_parts/mecha_equipment/tool/drill I believe). I did not try this with a diamond exosuit drill, hardsuit drill, or any other kind of mining tool. I did not attempt to drill "through" any obstructions except asteroid rock (eg walls, trash piles, windows, or tables).

Only noticed this today, current server revision. 

#### Issue bingo

https://puu.sh/HdsKa/f2549eafa1.png
Clicked in the middle of the 3x3 area. Note wall of asteroid rock still standing between me and the area mined.
https://puu.sh/HdsS2/56a00387ce.png
Clicked on wall where mouse cursor is, wall was drilled out as though it were beside me.
https://puu.sh/HdsPo/f0c68ec36c.png
https://puu.sh/HdsPw/a5332847b8.png
Before and after. Vision of the tile you click on is not necessary for the drill to operate.

zackerycodestoo:
## About The Pull Request

Fixes #206

Currently, mechs do not check if their target is in range when using the drill or the hydraulic clamp. It seems that a variable, inrange, was created and forgotten about. A check has been implemented and now mechs will only drill/clamp if they are actually adjacent to their target.

image
image
image
image

## Why It's Good For The Game

Quantum drilling and clamping is funny, but not intended

## Changelog
```changelog
fix: Mech drills and clamps no longer ignore range
```

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -14,54 +14,65 @@
 	. = ..()
 
 	if(. && !carrying)
-		if(istype(target, /obj))
+
+		///// OCCULUS EDIT START
+		// Mechs did not check if they were actually in range before trying to use drills or clamps.
+		// The proc has been adjusted with the following clause in mind.
+
+		if (!inrange)
+			to_chat(user, SPAN_NOTICE("You must be adjacent to [target] to use the hydraulic clamp."))
+		else
+
+			if(istype(target, /obj))
 
 
-			var/obj/O = target
-			if(O.buckled_mob)
-				return
-			if(locate(/mob/living) in O)
-				to_chat(user,"<span class='warning'>You can't load living things into the cargo compartment.</span>")
-				return
-
-			if(istype(target, /obj/structure/scrap))
-				owner.visible_message(SPAN_NOTICE("\The [owner] begins compressing \the [O] with \the [src]."))
-				if(do_after(owner, 20, O, 0, 1))
-					if(istype(O, /obj/structure/scrap))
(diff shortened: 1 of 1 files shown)


zackerycodestoo:
I'm aware the changelog still looks extremely screwy -- maybe it's something to do with the spacing in the file? All that's changed is that I added the if (!inrange) check to the start of both procs and then tabbed the rest.

Jamini (approved):
Yay gitkracken proving that it is inferior to bash once again!

PR #218, opened 2021-02-06
Shields start turned on, adds shield boost to power vote

zackerycodestoo:
## About The Pull Request

These are shield changes per the emojicracy at https://discord.com/channels/777375833084919839/777989657064636463/806045543369015328

Shields will now start the game turned on and a 'shield boost' proc has been added as a component of the 'Power the Ship' vote, with its name also changed (Power the Ship and Boost Shields.)

This does not change the shield settings, it only turns them on. An engineer's (or someone else's) intervention is still required for the following:

1. Making the shields self-sustaining
2. Adding atmospheric containment to the shields

image

**DME CHANGES**

This pull does include DME changes. Specifically:

1. A modular file has been added: zzzz_modular_occulus\code\controllers\subsystems\vote.dm
2. The old vote.dm has been disabled, which was at: zzz_modular_syzygy\vote.dm

I don't know if the DME needs to be adjusted serverside, but that's what's necessary here!

## Why It's Good For The Game

I've heard folks talking in Discord about how they'd really like to play, but they avoid doing so because there are no engineers around to turn on the shields. Now folks don't have to worry about meteors hitting the telecomms or the main hallway in the first ten minutes! By also including a vote to boost the shields, characters won't need to break in to adjust the shields. This may especially encourage people to play during lowpop times since the shields are up.

The downsides of this pull:

1. This is one less task for engineer players, but they can still configure the shields to be self-sustaining and have atmospheric containment.
2. We may see more 'power the ship' votes if there are no engineers around.

## Changelog
```changelog
tweak: The round begins with the shields turned on in their default settings.
add: A 'shield boost' component has been added to the 'Power the Ship' vote to bring shields to 100%.
```


Jamini:
We are currently on code freeze. Do Not Merge will be removed after the loot rework is merged.

Idea seems solid though.

Mr. Fox, we should work out some tasks for engineers to do throughout the round as a priority next.

Jamini (approved):
Everything modular or marked. Moving one file to the right location, and a change a lot of people have requested.

PR #220, opened 2021-02-06
Blattedin Fix

Jamini:
## About The Pull Request

Reagents can once again function on dead mobs. There is a flag in the reagents themselves that determines if it should or not.
Blattedin now revives roaches that come into contact with the blattedin clouds/smoke as originally intended.

## Why It's Good For The Game

Roach buffs! Specifically stuff that had been broken/bugged for a while!

## Changelog
```changelog
fix: chems now function on dead things. there is a flag that enables this. By default they do not.
tweak: Blattedin gas now respects the eris metabolism system. added on_touch proc to Blattedin. 
```


Michiyamenotehifunana (approved):
Everything checks out, non-modular bits are marked.

PR #224, opened 2021-02-08
Changes self-harm breakdown to explicitly set flags

zackerycodestoo:
## About The Pull Request

In a recent round, a player had an issue where they couldn't talk after the self-harm breakdown ended. It was discovered that their 'suppress_communication' flag was set to -1 rather than 0.

Inspecting the code, it seems that the self-harm breakdown, instead of setting 'suppress_communication' to 1 or 0, instead currently increments or decrements the flag with ++ and --. While this idea may be sound in theory, in practice, it's possible for the breakdown to conclude before it occurs, which left said player in a state where they couldn't communication between the flag was set to -1.

This change explicitly sets the communication suppression to 1 or 0 to avoid that situation.

## Why It's Good For The Game

Players who get lucky enough for a fast end to self-harm will no longer be at risk of permanent silence.

## Changelog
```changelog
fix: Self-harm breakdown can no longer accidentally cause permanent muteness
```

### DIFF
--- a/code/modules/sanity/breakdowns.dm
+++ b/code/modules/sanity/breakdowns.dm
@@ -203,11 +203,11 @@
 
 /datum/breakdown/negative/selfharm/occur()
 	spawn(delay)
-		holder.owner.suppress_communication = 1	// OCCULUS EDIT: was ++holder.owner.suppress_communication
+		holder.owner.suppress_communication = TRUE	// OCCULUS EDIT: was ++holder.owner.suppress_communication
 	return ..()
 
 /datum/breakdown/negative/selfharm/conclude()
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana (dismissed):
I'd prefer the usage of explicit `TRUE` and `FALSE`s rather than `0` and `1`s considering treating this thing as a number was what caused the issue to begin with, but it works nonetheless, and it's not like anyone's going to touch this code for the foreseeable future after this anyhow. Non-modular changes are also marked. Approved.

zackerycodestoo:
joke's on you I changed it to TRUE/FALSE

Michiyamenotehifunana (approved):
Re-approving.

PR #225, opened 2021-02-08
Rebalances gun and tool oldifiers to be in line with other oldifiers

Michiyamenotehifunana:
## About The Pull Request

This kinda counts as a buff, but not really because syzygy oldifier code nerfed maint guns and tools to utter uselessness. Now each stat has a *chance* of being reduced when oldified, rather than everything being guaranteed to be crappified. This is in line with the oldifier code of other items such as clothing. The chances of gun damage and tool durability being crapified is far higher than the other stats.

I have also opted to reduce the maximum reduction of tool workspeed and gun damage to 50% rather than 10%.

## Why It's Good For The Game

Maint guns should be a second-rate choice rather than literal clutter.

## Changelog
```changelog
balance: gun and tool oldifiers have been made less CBT.
```


Jamini:
1. The old code had a chance of not reducing stats as it could reduce stats by 0.
2. These probabilities are far too low. It should probably be around 60% for each stat to be reduced.
3. Maint guns *should* be garbage. We still find a lot of good guns in the trash (because not everything gets hit by the oldifier). Back when we had full-power guns in maint literally nobody bothered printing or buying guns. Now at least *some* weapons are bought and sold like they should be.
4. Using maint anything should be CBT.

bearrrrrrrr:
'it could have its stats reduced by 0' seems kind of, incredibly unlikely, and it literally makes them useless to the point where you shouldn't even bother. in between the new scrap ammo PR and kallin's new PR i really think this is fine imho??

Michiyamenotehifunana:
Oldification chance has been increased across the board, with a maximum of 80% chance for a gun's damage and a tool's health to be affected.

Jamini (approved):
Seems solid! Just modifying our modular files.

PR #232, opened 2021-02-09
Makes DNA properly transfer body marking stuff, fixes the winebath not applying body markings

Michiyamenotehifunana:
## About The Pull Request
Exactly what it says on the tin. I had copypasted some bits of DNA code that I did not fully understand, which turned out to be completely useless and dysfunctional. This is fixed. I have also made it so that the winebath will properly sync up organ DNA as they are regenerated, which also makes them properly apply body markings.

## Why It's Good For The Game

Less jank, more bugfixes.

## Changelog
```changelog Toriate
fix: fixed the winebath not properly applying body markings
```

### DIFF
--- a/code/game/dna/dna2_helpers.dm
+++ b/code/game/dna/dna2_helpers.dm
@@ -173,12 +173,15 @@
 		//Body markings
 		H.body_markings = dna.body_markings.Copy()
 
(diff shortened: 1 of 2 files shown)


zackerycodestoo (approved):
Non-modular changes marked, code functions as expected

PR #193, opened 2021-01-31
Ports multiple upstreams from Eris.

bearrrrrrrr:
## About The Pull Request

https://github.com/discordia-space/CEV-Eris/pull/5897
https://github.com/discordia-space/CEV-Eris/pull/5882
https://github.com/discordia-space/CEV-Eris/pull/5881
https://github.com/discordia-space/CEV-Eris/pull/5880

tl;dr adds trash ammo in maint for trash guns, as well as one-time-use kits that help you repackage/make ammo. fixes penetration, lets you dig out shrapnel and self-cauterize in big emergencies

## Why It's Good For The Game

We're changing scrap guns a bit. I think this + jamming would bring maint guns in line enough to allow them to have more consistent damage. If not, free mirror.

## Changelog
```changelog
add: scrap bullets/scrap bullet kits
```

PR #234, opened 2021-02-09
Revert "5881" Reverts overpenetration issues caused by bad Eris code

### LINKED ISSUE #233: Projectiles overpen to a stupid degree
#### Description of issue
Almost every bullet now since the scrap ammo update will act like a lethal rifle round when shooting roaches and not only kill the roach by fly off past it.

I don't expect shooting a .35 rubber out of a .8 pen gun at the ground to fly off and blast out windows or other people

#### Steps to reproduce
As above, just shoot any roach with any pistol and any round type and watch the bullets continue past the roach, when before they terminated at the roach.

Since the scrap ammo PR

zackerycodestoo:
This reverts commit 7b98b242329d938a42c63c5ec143333dd8dea780.

## About The Pull Request

Fixes #233

Bad Eris code as part of an upstream pull in #193 seems to be causing silly amounts of overpenetration due to bad math -- Kallin has indicated that he will try to salvage 5881 but currently it is going to be reverted due to all ballistic ammo penetrating (issue was reported while using a Paco with .35 rubber ammo -- testing the revert using the same weapon and ammo did not see the overpenetration issue)

## Why It's Good For The Game

Reverts to old penetration code -- rubbers will no longer pierce through walls and hit welding tanks/other fun things

## Changelog
```changelog
del: Undoes Eris #5881 which was causing overpenetration of bullets that should not have penetration
```

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -76,7 +76,7 @@
 	var/recentwield = 0 // to prevent spammage
 	var/proj_step_multiplier = 1
 	var/list/proj_damage_adjust = list() //What additional damage do we give to the bullet. Type(string) -> Amount(int)
-	var/noricochet = FALSE // wether or not bullets fired from this gun can ricochet off of walls
+
 /obj/item/weapon/gun/get_item_cost(export)
 	if(export)
 		return ..() * 0.5 //Guns should be sold in the player market.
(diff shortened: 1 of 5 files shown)


Michiyamenotehifunana:
Is it actually bad maths? Shouldn't we just set the penetration multiplier to 0 for all rubber bullets instead?

Jamini (approved):
Yes this is a bad issue. It's basically a saltPR for a gun that nobody uses that was implemented over on Eris because they have NO quality control.

Approved. Speedmerged.

PR #235, opened 2021-02-10
The Stungloves of the maintgoblins

RadiantFlash7:
## About The Pull Request
Adds in stungloves for all the ~greyshirts~ Maint gremlins!
Nearly identical to power gloves, only worse. takes 7 hits to stun compared to Aegis 4-5.
Has slightly more uses then Aegis power gloves, but requires more hits to incapacitate someone.
Craftable! Uses up a pair of insulated gloves, wires, power cell, and requires various tools to make.
Note: Craftable item recipes require to be in the same file to appear in menu due to code. Thus, Nonmodular.
stungloves

## Why It's Good For The Game
An alternative to stunprods, these stun gloves of old are ancient. Since Aegis has a set, I decided it'd be good to revive them in general, albiet worse then securities pair.

## Changelog
```changelog
add: stungloves
add: stungloves crafting
imageadd: added completed stunglove sprite(Old partial sprites were in code)
```


zzzz_modular_occulus/code/datum/craft/recipes/weapon.dm
```diff
@@ -0,0 +1,11 @@
+/*/datum/craft_recipe/weapon/stungloves
```
Michiyamenotehifunana, line 1: Why's this commented out!

code/datums/craft/recipes/weapon.dm
```diff
@@ -236,3 +236,16 @@
 		list(CRAFT_MATERIAL, 2, MATERIAL_GLASS, "time" = 10),
 		list(QUALITY_ADHESIVE, 15, 70)
 	)
+
+
+/datum/craft_recipe/weapon/stungloves //Occulus edit
```
Michiyamenotehifunana, line 241: Why's this not modular!

zzzz_modular_occulus/code/modules/clothing/gloves/stungloves.dm
```diff
@@ -0,0 +1,128 @@
+/obj/item/clothing/gloves/stunglovemakeshift //STUNGLOVES RETURN! MUAAHAHAHAHAHAHA!
```
Michiyamenotehifunana, line 1: Why's this entire thing copypasted instead of being a child object of the parent power gloves but with different stats!

Michiyamenotehifunana (commented):
There's a lotta copypasta in this! Are you sure you can't reduce it?

Also the onmob sprites have basically no indication that they're modified at all, so they'd be really, *really* stealthy unless you took the time to examine someone. I know it's a pain in the ass to do it, but maybe you could redden the onmob sprites a bit so it looks like it's got wires dangling all over it or something!

zzzz_modular_occulus/code/datum/craft/recipes/weapon.dm
RadiantFlash7: I originally tried making the recipe modular, but it’s not possible do so. I forgot to delete this file >..>

code/datums/craft/recipes/weapon.dm
RadiantFlash7: Because it can’t be modular. The code for the crafting recipe menu pulls datums from specific files to list them in the menu.

zzzz_modular_occulus/code/modules/clothing/gloves/stungloves.dm
RadiantFlash7: > There's a lotta copypasta in this! Are you sure you can't reduce it?
> 
Short answer: not really. The existing stungloves are not set up to have chilsren types. I originally tried to make to make it a child for it, but due to the way they handle sprites, it wouldn’t work.
My only options were to either change the core code file, which I prefer to avoid as much as possible to avoid future merge conflicts, or copy and paste the code for a new filepath of stungloves.
The stunglovesmaint /should/ be able to accept children types, I did adjust enough in the code for that.
> Also the onmob sprites have basically no indication that they're modified at all, so they'd be really, _really_ stealthy unless you took the time to examine someone. I know it's a pain in the ass to do it, but maybe you could redden the onmob sprites a bit so it looks like it's got wires dangling all over it or something!

I suppose I can do that! I kind of thought the semistealthy part was a bit of the appeal, hence why the stungloves are simply called modified insulated gloves, rather then makeshift stungloves.

RadiantFlash7:
> There's a lotta copypasta in this! Are you sure you can't reduce it?
>

Short answer not really.I originally went with the option of making it a child of aegises stungloves, but it wouldn’t work. The existing stungloves aren’t set up to have children types due to their coding for setting sprites.
 This means I’d either need to adjust the core eris files, or I’d need to copy paste the entire code and make changes to it myself.
I went with the latter due to trying to avoid touching files that would cause potential merge conflicts in the future.
> Also the onmob sprites have basically no indication that they're modified at all, so they'd be really, _really_ stealthy unless you took the time to examine someone. I know it's a pain in the ass to do it, but maybe you could redden the onmob sprites a bit so it looks like it's got wires dangling all over it 

To be honest, I kinda thought this was a feature. It’s why I named them modified Insulated gloves, rather then Makeshift Stungloves.
I /can/ make it more obvious if you really want though.

JMoldy:
oh no
not again

Jamini:
> oh no
> not again

Oh yes! :^)

code/datums/craft/recipes/weapon.dm
Jamini: > Because it can’t be modular. The code for the crafting recipe menu pulls datums from specific files to list them in the menu.

We have several unique crafting recipies that are modular. Disktoasters and sofas specifically. It should be perfectly possible to make this modular.

zzzz_modular_occulus/code/modules/clothing/gloves/stungloves.dm
Jamini: You can overwrite the update_icon() proc in the child file however. Most of this doesn't need to be copy-pasted.

RadiantFlash7:
I think, this should be litterally, all the problems with these solved. Pls rate and review. No criticism, only compliments, or whatever fanfic writers put in their things.

Michiyamenotehifunana (approved):
Copypasta's been drastically reduced. This seems like an okay compromise! All good.

PR #237, opened 2021-02-10
Prevent tracking of oil/blood/roachguts/etc if the source tile contains a catwalk or stairs

zackerycodestoo:
## About The Pull Request

The code that determines if you track footsteps after stepping in oil/blood/roachguts/etc has been modified to prevent any tracking if the turf containing the offending decal also contains a catwalk or stairs.

This is a non-modular change.

image

In the above image, the turf to my right contains both stairs and a gib that would normally cause tracks.

image

In the above image, I moved through the tile square -- no blood was tracked.

image

In the above image, I moved through a square that had gibs in it, and blood was tracked.

image

In the above image, robot gibs were present in the catwalk square to my left.

image

In the above image, oil was not tracked into the regular tiles because the new code prevented tracking since the source also contained a catwalk.

This code doesn't prevent gibs or blood from being spawned on tile or catwalk squares,

## Why It's Good For The Game

The reason for this change is that it's very difficult to spot and clean gibs that are on staircases and catwalks, so the player shouldn't be punished by making tracks everywhere or having gross shoes.

The benefits to this are twofold:

1. There will be significantly less cases of bloody/oily/whatever tracks coming out of maintenance tunnels/staircases
2. The (probably) low population of clean freaks (like me) don't have to worry about constantly cleaning their shoes after stepping on gibs they could not see in maintenance/staircases

## Changelog
```changelog
tweak: Footprints no longer track if the tracking source also contains stairs or a catwalk
```

### DIFF
--- a/code/game/objects/effects/decals/Cleanable/humans.dm
+++ b/code/game/objects/effects/decals/Cleanable/humans.dm
@@ -73,57 +73,50 @@ var/global/list/image/splatter_cache=list()
 	///// OCCULUS EDIT: Don't track blood if the source has a catwalk or stairs, since it is
 	//                  difficult to spot and clean.
 
-	var/do_not_track = FALSE
+	for (var/obj/structure/S in get_turf(perp))
+		if (istype(S, /obj/structure/catwalk) || istype(S, /obj/structure/multiz/stairs))
+			return
 
-	if (perp.loc.contents)
-		for(var/content in perp.loc.contents)
-			if(istype(content,/obj/structure/catwalk) || istype(content,/obj/structure/multiz/stairs))
-				do_not_track = TRUE
-				amount = 0
-				break
+	///// OCCULUS EDIT END
 
 	if(amount < 1)
 		return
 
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana (dismissed):
Dealing with catwalk/stairs gibs is a pain in the ass, so I guess this is a good QoL thing! I do wonder about the performance ramifications of iterating through the entire contents of a tile with a gib in it when a mob passes through though, so that might be a concern to keep in mind when we merge this.

zackerycodestoo:
I have a better idea for the iteration and will tackle this again later

zackerycodestoo:
@Michiyamenotehifunana I've adjusted the code that checks so it only iterates over obj/structure -- let me know what you think

Jamini (approved):
Wait, so the catwalks in maint ACTUALLY HAVE A PURPOSE NOW?

Approved. Wonderful work Mr. Fox

Jamini:
image

PR #245, opened 2021-02-11
Adjusts EoTP

Jamini:
## About The Pull Request

EoTP now has triple the chance of granting a stat boon instead of another reward.
Material rewards for the EoTP have been reduced to sane levels. You now get two smaller stacks of materials.
Moved EoTP to the bioreactor.
Mekhanite Seals are now 5000cr instead of 8000cr, in line with secret documents.
Mekhanite Seals now give BIO instead of COG
EoTP no longer can give uranium at all, as none of the church recipes require it.

## Why It's Good For The Game

The EoTP has needed a bit of a nerf for a while now. Now it's a bit less rediculous.

## Changelog
```changelog
balance: Nerfed EoTP
map: Moved EoTP to the bioreactor
balance: Reduced value on Mek seals
balance: Changed Mek seals to BIO instead of COG
```

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -2834,6 +2834,7 @@
 #include "zzzz_modular_occulus\code\modules\clothing\under\under.dm"
 #include "zzzz_modular_occulus\code\modules\clothing\under\jobs\medsci.dm"
 #include "zzzz_modular_occulus\code\modules\core_implant\cruciform\cruciform.dm"
+#include "zzzz_modular_occulus\code\modules\core_implant\cruciform\machinery\eotp.dm"
 #include "zzzz_modular_occulus\code\modules\core_implant\cruciform\rituals\base.dm"
 #include "zzzz_modular_occulus\code\modules\core_implant\cruciform\rituals\priest.dm"
 #include "zzzz_modular_occulus\code\modules\economy\power_cells.dm"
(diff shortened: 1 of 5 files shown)


cebutris:
Only thing I don't like about this is the increased chance for the stat buff. It already feels like a waste when it rolls, like you're missing out on something tangible. If you're making it more common, I'd like to see it at least improved in some way.

Other than that, everything else is good

Jamini:
> Only thing I don't like about this is the increased chance for the stat buff. It already feels like a waste when it rolls, like you're missing out on something tangible. If you're making it more common, I'd like to see it at least improved in some way.
> 
> Other than that, everything else is good

That is fair Cebu, however right now two other "No tangible reward" rolls (Antag radar and Positive breakdown) were removed by Tori with nothing put in place to replace them. That's partially why we are getting oh god materials.

I'd like to run it a bit and see how it feels for church members in an active church round or two.

cebutris:
> That is fair Cebu, however right now two other "No tangible reward" rolls (Antag radar and Positive breakdown) were removed by Tori with nothing put in place to replace them. That's partially why we are getting oh god materials.
> 
> I'd like to run it a bit and see how it feels for church members in an active church round or two

Understandable! 👍

zackerycodestoo (approved):
There's still the potential for a heck of a lot of materials (I know that 20 diamonds is more than what is needed for science, and 60 plasteel is probably enough to keep most departments going for an entire round) but I also know that it's a low chance of it happening with a higher chance for stat buffs. At least it won't be as bad as a church member showing up with 120 uranium, because who needs mining!

Non-modular changes are marked and modular files are present, all is well

Jamini:
image

PR #246, opened 2021-02-11
Adds autodoc machine

Jamini:
## About The Pull Request

This PR re-enables and fixes the autodoc machinery in code

Costs, directly identical to commercial hardsuit autodoc modules
#define AUTODOC_SCAN_COST           200
#define AUTODOC_DAMAGE_COST         800 (per 20 damage)
#define AUTODOC_EMBED_OBJECT_COST	1000 ((NOT CURRENTLY FUNCTIONAL. APPEARS BUGGED))
#define AUTODOC_FRACTURE_COST       1200
#define AUTODOC_IB_COST				1200 (per 10% blood)
#define AUTODOC_OPEN_WOUNDS_COST    600
#define AUTODOC_BLOOD_COST          800
#define AUTODOC_TOXIN_COST			600
#define AUTODOC_DIALYSIS_COST		1000

## Why It's Good For The Game

Adds a stationary autodoc in medical for lowpop rounds. The procedures are fairly expensive and quite slow, so it's difficult for anyone to simply spam. But in a pinch it can save you from certain death via a broken rib or the like. 

Functionally identical to a Commercial Hardsuit Autodoc.

## Changelog
```changelog
add: stationary autodoc in medical
```

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -2798,6 +2798,7 @@
 #include "zzzz_modular_occulus\code\game\jobs\job\medical.dm"
 #include "zzzz_modular_occulus\code\game\jobs\job\science.dm"
 #include "zzzz_modular_occulus\code\game\jobs\job\security.dm"
+#include "zzzz_modular_occulus\code\game\machinery\autodoc.dm"
 #include "zzzz_modular_occulus\code\game\machinery\jukebox.dm"
 #include "zzzz_modular_occulus\code\game\machinery\smartfridge.dm"
 #include "zzzz_modular_occulus\code\game\machinery\vending.dm"
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -78782,6 +78782,7 @@
 	pixel_y = 26
 	},
 /obj/machinery/camera/network/medbay,
+/obj/machinery/autodoc,
 /turf/simulated/floor/tiled/white,
 /area/eris/medical/reception)
 "dCd" = (
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -49732,7 +49732,9 @@
 /obj/machinery/door/firedoor,
 /obj/effect/window_lwall_spawn/smartspawn,
 /turf/simulated/floor/plating,
-/area/eris/medical/psych)
+/area/eris/medical/psych{
+	name = "\improper Autodoc Room"
+	})
 "clS" = (
 /obj/machinery/door/firedoor,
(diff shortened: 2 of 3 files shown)


Jamini:
image

Jamini:
image

Jamini:
After a GREAT DEAL of circular discussion, this passed emojicracy.

zackerycodestoo:
Entering the Autodoc as an FBP causes a runtime and makes the machine lock up after scanning your ID (runtime not present on an organic entering, I deleted and respawned the autodoc before doing so):

image
image

(The second image is a continuation of the first, the second image starts at the first's final line)

zackerycodestoo (changes requested):
Entering the Autodoc as an FBP causes a runtime and makes the machine lock up after scanning your ID (runtime not present on an organic entering, I deleted and respawned the autodoc before doing so):

image
image

(The second image is a continuation of the first, the second image starts at the first's final line)

zackerycodestoo (changes requested):
The area is marked as 'Psych Room' on GPS -- could that be adjusted to 'Autodoc Room'?

zackerycodestoo (commented):
Non-blocking comment: The Psychiatrist spawns in the Medical Isolator now, could that be moved to the Break Room?

Also, I understand that shrapnel removal is bugged -- I might suggest looking at the shrapnel removal code at code/modules/surgery/organic.dm, lines 410-417 -- the primary difference I see in the code here is that organic.dm refers to organs as _/obj/item/organ/external_ while the autodoc code is checking _/obj/item/organ/internal_

Jamini:
> Entering the Autodoc as an FBP causes a runtime and makes the machine lock up after scanning your ID (runtime not present on an organic entering, I deleted and respawned the autodoc before doing so):
> 
> image
> image
> 
> (The second image is a continuation of the first, the second image starts at the first's final line)
> 
> I know that FBPs shouldn't be able to use it but I'm a wild animal and wanted to see what would happen

The exact same thing probably would occur if a FBP tried to use an autodoc module on a hardsuit. It should be possible to clear this by leaving the autodoc (since when that occurs the old autodoc datum is deleted and a new one is created)

I'll move the psych spawn in the morning.

zackerycodestoo:
It did not clear the autodoc when I left it -- the runtime kept continuing every time I entered it after it scanned my ID (including possessing a normal human mob and entering it while it was in the lock-up state). The only way I was able to clear it was by deleting the autodoc and spawning a new machine

Jamini:
image

Could not reproduce the FBP thing.

Jamini:
Completely unable to reproduce

image

zackerycodestoo (approved):

PR #248, opened 2021-02-11
Smol lore update

Ikmachek:
Changed the NT bioprinter to be from Mekhane and pomme to be from FS


zackerycodestoo:
Please format like other PRs

PR #255, opened 2021-02-14
ports from Eris, again

bearrrrrrrr:
## About The Pull Request

Eris ports.
https://github.com/discordia-space/CEV-Eris/pull/5905
https://github.com/discordia-space/CEV-Eris/pull/5890

## Why It's Good For The Game

Free content. Adds some stuff a lot of people have suggested already, so...why not?!

- Add a Neat perk that Custodians have by default: You're used to see blood and filth in all its forms. Your motto: a clean ship is the first step to enlightenment. This perk reduces the total sanity damage you can take from what is happening around you. You can regain sanity by cleaning.

- Add a Green Thumb perk that Agrolytes have by default: After growing plants for years you have become a botanical expert. You can get all information about plants, from stats to harvest reagents, by examining them. Gathering plants relaxes you and thus restores sanity.

- Add a new Accelerated Growth littany exclusive to Agrolytes: This litany boosts the growth of all plants in sight for about ten minutes.

-Club personnel now starts with a trait thta has the same characteristics as the "charming personality" trait, giving passive sanity regen to those around them.

## Changelog
```changelog
add: Club worker and Mekhane people have new traits. New litany for agrolyte. 
```


bearrrrrrrr:
don't merge/review just yet pls, committed early like a dingus-

bearrrrrrrr:
good to go. seemed to run/work fine when i tested it but someone PLEASE look this over, i am terrified of spaghetti removed/readded/rebalanced/partially-modularized cruciform code

Jamini:
Please write what is added changed in detail, using the PR format Bear.

bearrrrrrrr:
oki

code/game/turfs/simulated/cleaning.dm
```diff
@@ -39,6 +39,10 @@
 		for(var/obj/effect/O in src)
 			if(istype(O,/obj/effect/decal/cleanable) || istype(O,/obj/effect/overlay))
 				qdel(O)
+		if(ishuman(user) && user.stats && user.stats.getPerk(/datum/perk/neat))
+			var/mob/living/carbon/human/H = user
+			if(H.sanity)
+				H.sanity.changeLevel(0.5)
```
zackerycodestoo, line 45: This code is flawed -- it doesn't activate when using cleaner spray, cleaner grenades, or even broad sweeps of mops, but only when you clean a single tile at a time (single-tile mopping, soap, wet rags, etc) and will likely significantly underwhelm any players expecting it to be a good source of sanity regain

zackerycodestoo: Additionally, it still triggers even if there's nothing to clean, so a custodian could theoretically spam their mop all over a clean floor to get their sanity back

code/datums/perks/job.dm
```diff
@@ -162,3 +162,47 @@
...
+	        to harvest reagents, by examining them. Gathering plants relaxes you and thus restores sanity."
+	icon_state = "greenthumb" // https://game-icons.net/1x1/delapouite/farmer.html
+
+	var/obj/item/device/scanner/plant/virtual_scanner = new
+
+/datum/perk/greenthumb/assign(mob/living/carbon/human/H)
+	..()
+	virtual_scanner.is_virtual = TRUE
+
+/datum/perk/job/club
```
zackerycodestoo, line 195: Perk appears non-functional -- I didn't see any change in a low-sanity mob's sanity gain rate from standing near it while holding this perk, both nearby and far away, besides the normal gain rate

zackerycodestoo (changes requested):
Bear, would you please look and see if Eris upstream has further pulls related to these? Here is what I found out:

1. The Club perk doesn't actually seem to have an effect, since the sanity gain rate of a low-sanity was unchanged while I was near them with the perk
2. The Custodian perk is significantly underwhelming and does not trigger on spray bottles, cleaner grenades, or broad sweeps of the mop -- only on single-target cleaning, and _also_ can be triggered by cleaning the same clean tile repeatedly

The Agrolyte perks and ritual at least seem to work as expected.

Other notes I made and confirmed:

1. /obj/item/device/scanner and its children still work as expected
2. The Agrolyte examine still gets 'print report' and 'clear data', but those buttons don't do anything, so it's fine
3. Plants still grow without boosted growth (I was mighty suspicious of this one)

bearrrrrrrr:
I specifically double- and triple- checked, there are no fixes available with these changes. Up to yall if you don't want to get the new traits since they're sorta buggy or just want the new litany or want to close this entirely.

zackerycodestoo:
I do like the Agrolyte changes -- I would say to drop 5890 (Club Worker aura) from this pull and I can try to help you drop the Neat perk. The only reason I say to drop them is that it would cause folks to think they're a lot more effective than they actually are (and in the Club Worker perk's case, it doesn't work at all)

bearrrrrrrr:
Sure, can-do.

bearrrrrrrr:
These changes good? Removes club perk, clarifies Filthy perk. Tell me if you want the jannie one removed entirely.

zackerycodestoo (approved):
Thanks! It's fine to merge now that the text has been clarified

PR #256, opened 2021-02-14
Fixes code to check for coverage while injecting

### LINKED ISSUE #40: Syringe Injection code does not check armor coverage to prevent needles
#### Description of issue

Injection code should check part coverage for preventing injection

Currently all armor blocks injection, even armor that doesn't cover limbs.

#### Steps to reproduce 

1. Put on armor
2. Pick up a needle
3. Fill needle
4. Attempt to inject self

This has been the case as long as Occulus and Syzygy has existed. Eris likely did not port changes that were made to the bay branches when they split

#### Issue bingo

zackerycodestoo:
## About The Pull Request

Fixes #40

Currently, any armor with the THICKMATERIAL flag worn on the suit slot will prevent syringe injections into any part of the body. This is due to the code only checking the worn suit slot for THICKMATERIAL rather than the body part you're trying to inject into.

This was solved by hijacking some code that determines armor coverage and modifying it to test for THICKMATERIAL presence. If there's no THICKMATERIAL flag, you can inject!

Testing log:

Test subject: Wearing /obj/item/clothing/suit/armor/vest (armor vest with THICKMATERIAL tag, covers upper and lower torso)

image

1. Injection site: Upper torso -- 
image
2. Injection site: Lower torso -- 
image
3. Injection site: Right arm -- 
image
4. Injection site: Left foot -- 
image

The code that checks if the head is covered was left untouched. This code adjustment does not impact implanting, since that is already handled separately and performing as expected.

This code contains non-modular changes.

Thanks to Leon9621 for the clue on where to look to fix this.

## Why It's Good For The Game

You can now shoot up needles into your Aegis friends' exposed limbs without pulling their vests off first!

## Changelog
```changelog
fix: Thick armor no longer blocks injections on non-covered limbs
```

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1361,13 +1361,27 @@ var/list/rank_prefix = list(\
 				if(head && head.item_flags & THICKMATERIAL)
 					. = 0
 			else
-				if(wear_suit && wear_suit.item_flags & THICKMATERIAL)
-					. = 0
+
+				///// OCCULUS EDIT START
+				// Fix injection code to not just check the 'wear_suit' variable
+				// Code partially hijacked from the armor protection methods
+
+				var/list/protective_gear = list(wear_suit, w_uniform, gloves, shoes)
+
+				for(var/gear in protective_gear)
+					if(gear && istype(gear ,/obj/item/clothing))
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Looks alright from this end.

PR #260, opened 2021-02-16
Smol item name change, smol breakdown QoL fix

Ikmachek:
## About The Pull Request
Change the Pomme and bioprinter to reflect our lore, modular. Changes Herald to whisper your madness and changes obsession to 5 seconds so other breakdowns aren't so common and you still get the insight bonus.

## Why It's Good For The Game
The lore changes explain themself but these two breakdowns have been on the nerves of players for MONTHS

## Changelog
```changelog
tweak: lore-reflects the pomme and bioprinter
fix: makes Herald and Obsession breakdown more tolerable
```


zzzz_modular_occulus/code/game/occulus-lore-fixes.dm
```diff
@@ -193,8 +193,6 @@
 	name = "Kampfer Roach"
 	desc = "A monstrous, dog-sized cockroach. These huge mutants can be found everywhere on the ship."
 
-/obj/item/weapon/tool_upgrade/augment/randomizer
```
zackerycodestoo, line 188: Is this removal intended? It would revert randomizer name, rarity, and give it the loot spawn tag back

zackerycodestoo (changes requested):
In the future, please don't lump unrelated changes into a single pull (as in, the lore updates and the breakdown tweaks should be separate pulls, so that both wouldn't get affected in the unlikely event of a revert)

Pull request content looks fine -- non-modular changes are marked

Is the removal of the modular overwrite of the randomizer paint/addition of drone flavor text intended, since I don't see mention of them in the text? That is the only blocker currently

zackerycodestoo (approved):
Fine now!

PR #262, opened 2021-02-17
Makes Glassification less silly

Michiyamenotehifunana:
## About The Pull Request

Now it only lasts for precisely one second, and is only two way if the recipient's sanity is low enough.

Port of https://github.com/discordia-space/CEV-Eris/pull/5332

## Why It's Good For The Game
Makes Glassification far less obnoxious and immersion breaking.

## Changelog
```changelog Toriate
add: Glassification is now only bidirectional if the recipient happens to be at low sanity. Also lasts for one second only now.
```


Jamini (approved):
Nothing unusual here. Decent port. Should make Glassification less obnoxious

PR #239, opened 2021-02-10
Burn decal cobwebs with welding/cauterizing tools (like lighters)

zackerycodestoo:
## About The Pull Request

This pull adds modular code to allow the decal cobwebs to be burned away with tools that have cauterizing (lighters) or welding qualities. This does not touch the cobwebs created by giant spiders.

Modular change: zzzz_modular_occulus/code/game/objects/effects/decals/Cleanable/misc_occulus.dm must be included

image
image
image
image

I notice that cobweb and cobweb2 are different objects -- later on I will try to learn what I need to merge them together 

## Why It's Good For The Game

You can now burn away cobwebs with a lighter or welder or other tool with cauterizing/welding.

I ran into a cobweb on the server and I tried to burn it away with my Zippo. I could not. My immersion was broken

## Changelog
```changelog
add: Decal cobwebs can now be burnt away with cauterizing/welding tools (like lighters)
```

PR #271, opened 2021-02-19
Activates #239 (burn cobwebs)

zackerycodestoo:
## About The Pull Request

Pull #239 was not actually active because I forgot to include DME changes in the original pull. This activates the code.

## Why It's Good For The Game

Now you can burn decal cobwebs (big spider ones are untouched) with welding/cauterizing tools (like lighters

## Changelog
```changelog
fix: Cobweb burning with welding/cauterizing tools (lighters) wasn't actually active, whoopsy doopsy
```


Jamini (approved):
Well that explains a lot.

PR #280, opened 2021-02-22
Adjusted wound processing

Jamini:
## About The Pull Request

Organs with wounds now continue to update them until the wounds are gone
Wounds now go away after being at 0 damage, salved, and bandaged instead of vanishing as soon as they hit 0 damage after ten minutes from the cause.
Wounds now process every third tick instead of every tick

## Why It's Good For The Game

It is now possible to remove a page worth of old injuries by treating them properly

## Changelog
```changelog
tweak: Wounds now process every third tick
tweak: Wounds that are fully healed, bandaged, and salved will no longer show on examine
```

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -44,7 +44,7 @@
 	var/skin_tone			// Skin tone.
 
 	// Wound and structural data.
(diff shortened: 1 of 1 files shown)


zackerycodestoo (approved):
Works as expected. There is a negligible delay on the wounds disappearing when damage hits 0 -- this is expected for the slowed tick checking. Lovely!

PR #283, opened 2021-02-24
Gun modularity

JMoldy:
## About The Pull Request
Allows firearms to change the knockback effect of bullets they fire.

## Why It's Good For The Game
Modularization of firearms, begins parts of firearms backend changes

## Changelog
```changelog
add:Firearms can now change their projectiles' knockback
```


zzzz_modular_occulus/code/modules/projectiles/projectile/projectile.dm
```diff
@@ -1,6 +1,9 @@
+/obj/item/projectile/change_knockback(change)
+	knockback = initial(knockback) + change
```
zackerycodestoo, line 2: Does this decay over time or does knockback continue building over multiple gunshots (not in rapid succession?)

zzzz_modular_occulus/code/modules/projectiles/gun.dm
```diff
@@ -37,4 +40,73 @@
...
+
+		projectile.multiply_projectile_damage(damage_multiplier)
+
+		projectile.multiply_projectile_penetration(penetration_multiplier)
+
+		projectile.multiply_pierce_penetration(pierce_multiplier)
+
+		projectile.multiply_projectile_step_delay(proj_step_multiplier)
+
+		projectile.change_knockback(knockback_change)
```
zackerycodestoo, line 79: Is this code any different from existing projectile code with the exception of adding the call to change_knockback()?

zackerycodestoo (changes requested):
Some questions to answer, please see comments -- I am mostly concerned that knockback does not get cleared properly, does it have an effect at all right now?

zzzz_modular_occulus/code/modules/projectiles/projectile/projectile.dm
JMoldy: I am not sure I understand the question here. If a projectile has knockback and it is changed to 0 then it simply acts as a projectile that has knockback = 0 and if it is changed from 0 to 1 it now gives it knockback.

zzzz_modular_occulus/code/modules/projectiles/gun.dm
JMoldy: This is functionally the same as the way existing code works.

However I am going to add a sanity check on this that isn't needed on other code.

zackerycodestoo (approved):

PR #291, opened 2021-02-26
Open surgery UI when clicking an open organ with help intent

zackerycodestoo:
## About The Pull Request

This is non-modular code that opens the surgery UI if you click the open organ with an empty hand while on help intent.

The purpose of this code is twofold:

1. If you accidentally close the surgery UI, you no longer have to close and re-open the organ to bring the UI back up.
2. Multiple surgeons can now interact in a surgery.

There is also an addition of a variable called 'playsound' that determines if the sound for shaking/hugging/whatever is played. This is TRUE by default and is only set to false if you are clicking on the open organ with help intent.

## Why It's Good For The Game

It's possible for you to be in a situation where it's good for more than one surgeon to be working on a patient. Specifically, this will allow a situation like a roboticist to be able to insert augmentations after a separate surgeon has opened a patient without needing to close/re-open the incision themself.

## Changelog
```changelog
add: The Surgery UI can now be opened by clicking on the open body part with an empty hand on help intent.
```

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -208,7 +208,11 @@
 				show_ssd = H.species.show_ssd
 				var/obj/item/organ/external/O = H.get_organ(M.targeted_organ)
 				target_organ_exists = (O && O.is_usable())
-			if(show_ssd && !client && !teleop)
+			// OCCULUS EDIT START: Open the surgery UI if the organ is open by clicking the organ.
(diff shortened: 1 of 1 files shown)


zackerycodestoo:
Concerns raised in Discord that this would prevent CPR -- it does not prevent CPR, but a situation where you can do CPR will override opening the surgery UI. Marked as WIP and will try changing this to a verb instead

zackerycodestoo:
Closing for now so it's not sitting in the PR list

PR #303, opened 2021-03-02
Smelters no longer qdel mobs being processed inside of containers

Jamini:
## About The Pull Request

Smelters no longer incidentally qdel mobs inside
Instead any mobs in a container that gets processed take 80 brute and 20 burn and are ejected.
Any equipment they are wearing is processed normally
Synthetic limbs are severed and ejected with the body

## Why It's Good For The Game

This is a fix for a silly bug, and now you can use some traitor shenanigans with recyclers!

## Changelog
```changelog
fix: mobs no longer get qdelled by smelters if they are in a locker
```


code/game/machinery/smelter.dm
```diff
@@ -128,6 +127,20 @@
 	for(var/obj/O in smelting.contents)
 		smelt_item(O)
 
+	//Occulus Edit Start: Smelting Mobs
+	for(var/mob/oopsalltori in smelting.contents)
```
Michiyamenotehifunana, line 131: *scream

Michiyamenotehifunana: How does it work with roaches and simplemobs and stuff if it doesn't sanitize down to the human type tho?

Jamini: take_overall_damage is a proc on mob/living
the organ separating proc only checks contents for external organs
the drop_all_from_inventory is only called if an item is in the contents of a mob and is flagged as equipped.

Jamini: Also I did test it with roaches and simplemobs. Most of them gibbed from the high damage, but Fuhrer's survive the process.

Michiyamenotehifunana (approved):
Everything seems about right, non modular changes marked. Should be good.

code/game/machinery/smelter.dm
zackerycodestoo, line 131: Fun police here -- I do think the in-joke is funny but future coders looking at this code might be confused because 'oopsalltori' and 'oopsallzack' aren't clear on a glance at what they refer to. This is a similar thing to how we have a variable called 'howdoitameahorseinminecraft' that I strongly dislike and intend to change because there's no clarity on what it does from a glance. I would ask you please change the variable names to something more descriptive (like `smelting_mob` and `external_organ`, so line 140 reads as `for(var/obj/item/organ/external/external_organ in smelting_mob.contents) ` or another similar descriptive variable)

zackerycodestoo (dismissed):
Please see my comment on more descriptive variable names

Jamini:
> Please see my comment on more descriptive variable names

Fuck the police. :frowning_face:

Michiyamenotehifunana (approved):
Variables are now understandable, looks good.

Issue #176, opened 2021-01-26
Mining smelter does not process making plasteel correctly.


SectoidDan:
#### Description of issue
Plasteel is an alloy of hematite (iron), carbon (coal), and phoron. Steel is an alloy of iron and carbon. When the smelter is set to smelt iron, carbon, and phoron, it produces both plasteel and steel sheets at the same time.

#### Difference between expected and actual behavior
Ultimately, the smelter should be configurable to output only plasteel and nothing else.
When the smelter is set to smelt iron, carbon, and phoron all together, it should produce only plasteel and no steel. Alternatively, the recipe for plasteel should be changed to not overlap with steel, perhaps by simply removing carbon from the alloy recipe. 

Note also that plasteel already has a 0.3 production modifier from alloys.dm - this bug compounds the difficulty of producing plasteel for miners and cargo in general beyond the intended level.

#### Steps to reproduce 
1. Put hematite, carbon, and phoron ore in a smelter. 
2. Set it to smelt all three. 
3. Watch and cry as most of the ores get turned into useless cheap steel instead of glorious strong (and most importantly) valuable plasteel.

#### Specific information for locating

code/modules/mining/alloys.dm and code/game/machinery/smelter.dm are probably the relevant files.

#### Length of time in which bug has been known to occur

Since I started playing here at least.

#### Client version, Server revision & Game ID

master - 2021-01-24

#### Issue bingo

Please check whatever applies. More checkboxes checked increase your chances of the issue being looked at sooner.

- [✔] Issue could be reproduced at least once (all the time)
- [✔] Issue could be reproduced by different players (I asked Rynn but it happens literally every time anyway)
- [✔] Issue could be reproduced in multiple rounds (every round)
- [✔] Issue happened in a recent (less than 7 days ago) round (yes)
- [✔] [Couldn't find an existing issue about this](https://github.com/discordia-space/CEV-Eris/issues)

zackerycodestoo:
Some inspection of the code seems to reveal that this is working as intended in the way that it tries to smelt plasteel first and then only smelts steel if it runs out of ingredients. However, I do agree that it would be much, much better if you could choose exactly what gets alloyed if you aren't interested in smelting steel by itself.

[...]

PR #314, opened 2021-03-06
Fixes mineral processors processing multiple alloys at once

Jamini:
## About The Pull Request

Multiple alloys can no longer be smelted simultaniously

## Why It's Good For The Game

This prevents alloys that share a portion of their recipe (specifically plasteel and steel) from processing at the same time.
Bumped Plasteel effeciency to 0.4

## Changelog
```changelog
fix: smelters no longer run every possible alloy recipe at once.
```

### DIFF
--- a/code/modules/mining/alloys.dm
+++ b/code/modules/mining/alloys.dm
@@ -15,7 +15,7 @@
 		"carbon" = 2,
 		"hematite" = 2
 		)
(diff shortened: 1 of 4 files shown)


Jamini:
fixes:  #176

zackerycodestoo (approved):
Much more elegant solution

PR #322, opened 2021-03-07
Prosthetics fabricator can print stomachs now + small clarity tweaks

zackerycodestoo:
## About The Pull Request

This non-modular code accomplishes the following:

1. All organ fabricators (both Church and robotics' prosthetics fabricator) can now print stomachs.
2. Synthetic organs are now clearly marked as synthetic to avoid confusion.
3. The fabricator output message is more clear about what's going on.

image
image

## Why It's Good For The Game

In a recent round, there was an issue where a stomach went septic from necrosis. Robotics was asked to print a new stomach, but it was discovered that stomachs couldn't actually be printed. There was also momentary confusion about which new organ (in the case of also needing lungs) was synthetic.

## Changelog
```changelog
add: Organ fabricators (organic and prosthetic) can now print stomachs.
tweak: Synthetic organs are now more clearly marked in case you find any in the trash.
tweak: Organ fabricators are a little more clear in their production messages.
```


Jamini (approved):
Looks good.

PR #334, opened 2021-03-14
Autodoc: Skip empty operations

### LINKED ISSUE #259: Autodoc needs performance improvements
#### Description of issue

In a recent round, a player had broken ribs. They used the autodoc and tried to fix their bones, but the Autodoc went into a spam of 'treatment complete' instead of actually fixing anything. It still ate their money.

This information is secondhand from the player who suffered the broken ribs.

#### Steps to reproduce 

1. Have broken ribs.
2. Try to fix broken ribs in the autodoc.
3. Autodoc spams 'treatment complete'

The autodoc currently in medbay

Since the autodoc machine was implemented

image

#### Issue bingo

- [ ] Issue could be reproduced at least once

zackerycodestoo:
## About The Pull Request

Fixes #259 

This non-modular code improves the autodoc code so that it skips empty operations rather than making the user wait for it to process before saying 'oh, nothing to do!'

This is not the most ideal way to do it -- ideally, the code would be rewritten in a way that notes not in the `picked_patchnotes` variable do not cause the UI to explode. Unfortunately, excluding empty patchnotes results in the autodoc UI becoming unresponsive. I am not skilled enough to disentangle the code to rewrite it, since it also involves .tmpl files.

Nonetheless, this code no longer makes players wait for multiple treatments when they have only selected one.

Finally, this also fixes a runtime in the autodoc code where it would try to check for a non-existent surgery operation (due to faulty code logic.) The runtime is harmless as the code was terminating anyway but it still fixes it nonetheless.

Thank you very much to Flipp for the thorough investigation into why the code was misbehaving.

## Why It's Good For The Game

On the player side of things, you are no longer waiting a very long time if you want to fix only a certain thing, which is the case for many deckhands/vagabonds using the machine. If you have multiple injuries and you only want to fix your ribs, this will only fix your ribs instead of iterating over every other possible treatment for 30 seconds each time and _then_ saying there's nothing to fix.

## Changelog
```changelog
fix: Autodoc will no longer attempt (empty) operations that you haven't selected.
```

### DIFF
--- a/code/modules/surgery/autodoc.dm
+++ b/code/modules/surgery/autodoc.dm
@@ -186,14 +186,32 @@
 /datum/autodoc/Process()
 	if(!patient)
 		stop()
+
+
+	/* OCCULUS EDIT START: Skip empty patchnotes. This is not the most ideal
+	   solution -- ideally, picked_patchnotes would not be populated with
+	   empty notes, but unfortunately the code is unintuitive and not
+	   programmed this way. I am not yet skilled enough to disentangle the
+	   code to rewrite it properly. */
+
 	if(current_step > picked_patchnotes.len)
 		stop()
+		to_chat(patient, SPAN_NOTICE("Operations complete."))	// OCCULUS EDIT: Tell the patient when it's over
 		scan_user(patient)
-	if(world.time > (start_op_time + processing_speed))
-		start_op_time = world.time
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Seems to work as expected.

PR #335, opened 2021-03-14
Allow .40 ammo to stack to 15 casings

### LINKED ISSUE #221: 40 ammo doesn't stack as much.
#### Description of issue
Handfuls of .40 ammo only stack up to 6 rounds now, compared to the 15 they previously could be stacked to. Could be the same with the other calibers.

Before it was space efficient, if not quick for mag fed .40 weapons, to store them as loose rounds, especially if you were using a .40 revolver or lever action, considering the odd numbers for their cylinders.

#### Steps to reproduce 
Get .40 magazine, ammo pouch, or speedloader, take out ammo.

.40 ammo

Since the ammo sprite update

master - 2021-02-06

#### Issue bingo

zackerycodestoo:
## About The Pull Request

Fixes #221 

This modular code allows .40 ammo to stack to 15, as they were previously reported to be able to do so. It may be that the original coder that reduced them to 6 had only revolvers in mind.

I do notice that other bullets are also limited to less than 15, but this may be for balance reasons, like shotgun shells and anti-materiel rifles. Allowing other ammo types to stack up to 15 might be a topic for a small discussion.

image

## Why It's Good For The Game

This saves .40 ammo users some headache when trying to organize their loose ammo.

## Changelog
```changelog
tweak: .40 ammo can stack to 15 casings, up from 6
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/projectiles/ammunition/bullets.dm
+++ b/zzzz_modular_occulus/code/modules/projectiles/ammunition/bullets.dm
@@ -2,3 +2,6 @@
 	icon = 'zzzz_modular_occulus/icons/obj/ammo.dmi'
 	icon_state = "75"
 	spent_icon = "75-spent"
+
+/obj/item/ammo_casing/magnum
+	maxamount = 15


Jamini (approved):
Nice modular override. nothing unexpected.

We need to look into why sometimes our bullets and casings have no sprite.

PR #337, opened 2021-03-15
Make superior animals (roaches, spiders) obey gravity

### LINKED ISSUE #269: Roaches do not follow the laws of the lack of gravity.
#### Description of issue

Roaches move normally in zero gravity.

Expected: Meet roach swarm. Go "Oh fuck" and close door. Get stumbled around, recover, escape from roaches who all have to move in 0G like you do.

Actual: Roaches move like normal and can stumble you, effectively stunlocking you forever to death. Because they will always be on top of you due to 0G "slip" floatiness.

#### Steps to reproduce 

No Grav. Big Room. Roach Mob.

18:20 18/02/2021

Server revision: master - 2021-02-16
5bee4f43afbd0c5302ad45c5482d7ac12827423e

#### Issue bingo

zackerycodestoo:
## About The Pull Request

Fixes #269

This non-modular code adds checks for superior animal movement to make them obey gravity. The issue is that the superior animal's movement code utilizes the BYOND proc `walk_to` which _continuously_ moves the mob towards the target, only stopping when they lose sight or when something gets in the way. The most ideal solution would be to write code that replicates `walk_to` but has in-built respect for gravity, though that may be beyond the scope of my skill at this time.

Instead, the code now has a few checks for gravity and then triggers `walk_to`.

## Why It's Good For The Game

Roaches and spiders will no longer see gravity and dismiss it as a social construct, so it should be easier to evade them in bad situations.

## Changelog
```changelog
fix: Superior animals (roaches, spiders, etc) now obey gravity.
```

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/life.dm
+++ b/code/modules/mob/living/carbon/superior_animal/life.dm
@@ -20,6 +20,9 @@
 	if (!check_AI_act())
 		return
 
+	if (!check_gravity() && !allow_spacemove())	// OCCULUS EDIT: if there is no gravity and spacemove is disallowed...
+		walk(src, 0)							// 				 stop movement.
+
 	switch(stance)
 		if(HOSTILE_STANCE_IDLE)
 			if (!busy) // if not busy with a special task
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Seems good.

PR #341, opened 2021-03-18
Teleport module nerf

Jamini:
## About The Pull Request

Blacklist teleport modules from random spawns
Teleport modules now cost 2,000 cell power per use.

## Why It's Good For The Game

This particular module has been problematic for a while. It should be much rarer/more expensive to acquire now, and have a significantly reduced number of jumps.

## Changelog
```changelog
balance: teleport modules can no longer be found in maint pools
balance: teleport modules now cost 2,000 energy, up from 40.
```


zackerycodestoo (approved):
Balance concerns cited -- code looks well

PR #345, opened 2021-03-20
fixed two up pipes that should be trunks

Jamini:
## About The Pull Request

Tiny map error fixed 

## Why It's Good For The Game

whoop

## Changelog
```changelog
fix: Fixed two up disposals that should be trunks
```


zackerycodestoo (approved):
All's well

PR #350, opened 2021-03-21
Update human_attackhand.dm

Jamini:
## About The Pull Request

Minor non-modular tweak. TGH now makes you harder to shove. 1:1 with ROB making someone more likely to shove you.

## Why It's Good For The Game

This makes it harder to disarmspam people with high TGH, and reduces shove chance overall

## Changelog
```changelog
balance: chance of stun on disarm now scale with target TGH as well as attacker ROB
```

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -270,7 +270,7 @@
 						return W.afterattack(target,src)
 
 			var/randn = rand(1, 100)
-			randn = max(1, randn - H.stats.getStat(STAT_ROB) + H.stats.getStat(STAT_TGH))//Occulus Edit: TGH makes you harder to shove
+			randn = max(1, randn - H.stats.getStat(STAT_ROB) + src.stats.getStat(STAT_TGH))//Occulus Edit: TGH makes you harder to shove
 			if(!(species.flags & NO_SLIP) && randn <= 20)
 				apply_effect(3, WEAKEN, getarmor(affecting, ARMOR_MELEE))
 				playsound(loc, 'sound/weapons/thudswoosh.ogg', 50, 1, -1)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -270,7 +270,7 @@
 						return W.afterattack(target,src)
 
 			var/randn = rand(1, 100)
-			randn = max(1, randn - H.stats.getStat(STAT_ROB))
+			randn = max(1, randn - H.stats.getStat(STAT_ROB) + src.stats.getStat(STAT_TGH))//Occulus Edit: TGH makes you harder to shove
```
zackerycodestoo, line 273: BYOND evaluates addition before subtraction (ref 10.5 https://secure.byond.com/docs/guide/chap06.html) -- in a theoretical fight of 50 ROB vs 50 TGH and a randn var of 33, 50 will be added to 50 for 100, and then 33 - 100 = -67, which will result in the `max` selecting 1 and guaranteeing a knockdown.

Please encapsulate the subtraction to make it evaluate first -- so in that same theoretical `randn` 33 w/ ROB 50 vs TGH 50, it becomes (33 - 50) + 50 for a final value of 33.

zackerycodestoo (changes requested):
Small adjustment required to formula, see comment

zackerycodestoo (approved):
2AM brain misread, all is well

code/modules/mob/living/carbon/human/human_attackhand.dm
Jamini: Figure 6.10 summarizes the order of operations of the mathematical symbols from highest to lowest. Operators on the same line have equal precedence and are therefore evaluated as they appear in the expression from left to right. 


( ) ++ -- -(negation)
--
**
* / %
+ -

PR #353, opened 2021-03-23
Cleanbot lagfix & associated changes

zackerycodestoo:
## About The Pull Request

This modular code fixes the Cleanbot's code to be less lag-heavy. 

With the original code, the mob controller would often see a deviation of +~200ms when there was an active cleanbot -- and that was just for one cleanbot. To frame it a bit, the average mob processing time (due to number of roaches, etc) is 300-400ms -- with a single cleanbot, that bumped it up to 500-600ms.

The culprit appears to be that the function that searches for tiles to clean was being ran every tick -- I'm not sure what the tickrate of the server is, but I assume that means the search was running dozens times _every second_. To combat this, the following changes have been made:

1. The cleanbot now has an 'active state' with a 300ms delay and a 'dormant state'. 

In the active state, the average mob delay increases by about 60ms rather than 200ms, and additional cleanbots beyond one do an average of +20ms each. If the cleanbot cannot reach any cleanables or can't find any cleanables, it will enter a dormant state.

In the dormant state, the cleanbot will only process every 30 seconds. If a cleanable that is NOT on its ignorelist (like a cleanable it couldn't reach before, or gibs spawned by a malfunctioning bot), it will enter an active state and begin cleaning until it meets the conditions detailed above, then it will return to a dormant state.

Finally, there has been a 'manual scan' button added to the cleanbot interface -- by pressing this button, you can force a scan, which clears the ignorelist and looks for cleanables again (in the event of a cleanable that was inaccessible before, or if a cleanbot is dormant when a lot of gibs suddenly appear.) In the future I would like to add voice activation.

There are now flavour messages, names, and descriptions for dormant and active cleanbots.

**High-priority has been assigned due to the Discord server being pinged that it is an issue.**

## Why It's Good For The Game

Now cleanbots can be made without painfully slowing the server down. There will be a bit of a spike while cleanbots work on a large area of cleaning, but not nearly as significant as before.

## Changelog
```changelog
fix: Cleanbots no longer severely lag the server.
add: Cleanbots now have active and dormant states and have a button in their UI to force a manual scan (if they are in a dormant state)
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/mob/living/bot/cleanbot_occulus.dm
+++ b/zzzz_modular_occulus/code/modules/mob/living/bot/cleanbot_occulus.dm
@@ -1,11 +1,21 @@
+#define STATE_DORMANT (1<<0)
+#define STATE_ACTIVE  (1<<1)
+
 /mob/living/bot/cleanbot
+
 	name = "Cleanbot"
+	var/active_name = "Cleanbot"
+	var/dormant_name = "dormant Cleanbot"
+
 	desc = "A little cleaning robot, he looks so excited!"
+	var/active_desc = "A little cleaning robot, it looks so excited!"
+	var/dormant_desc = "A little cleaning robot. Its LEDs are pulsing in a low-power mode."
+
 	icon_state = "cleanbot0"
 	req_one_access = list(access_janitor, access_robotics)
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Let's see how this goes.

PR #361, opened 2021-03-31
Hypospray rework 2.0

RadiantFlash7:
## About The Pull Request

The long awaited hypospray rework.
Reworks Hyposprays into Medical Injection Device Apollo.
The MID Apollo takes 45u cartridges,
scans a target for it's reagents, displaying it on a toggleable displayer,
Prevents overdoses upon use, notifying user of OD.
cartridges dispenses in 5, 10, and 15 u doses.
When emagged, it bypasses the overdose check, and injects instantly, similar to current hyposprays.
Adds variants to the MID Apollo known as the 'Hecate' and the 'Ares' Respectably.
MID Ares is the pre-emagged combat version for traitors, and can combatswap cartridges like a pistol in a much smoother fashion. The traitor gear also comes with combat stims.

The Hecate is the CMO variant, and the 'original' MID found on the northernlight, which Nanotrasen bases the Apollo copy off of. It is faster at injecting then the normal Apollo, and takes larger sized cartridges.
image
image

~~This PR also adds 2 new chemicals,
Ambursidol: A mild burn medication, currently stocked in burn MID kits, and can be made with innaprovaline and hydrochlorica acid. This heals at a rate in between Tricord and kelotane.
Lacertidam: A mild brute healing medication, currently stocked in brute MID kits, can be made with innaprovaline and aluminum. This heals at a rate in between Tricord and bicaridine.~~
PR now instead buffs Poly stem to a rate between tricord and bicaridine. Polystem is used in brute hypokits, Kelotane is used in burn hypokits.

## Why It's Good For The Game
Hopefully, gets the hypospray to a good place, provides some bonuses to the slowness offset, and makes the fact that the injection is slower then syringes, actually make sense due to scanning. Optional reagent scan is also very useful.

## Changelog
```changelog
~~add: Ambursidol and Lacertidam chemicals~~
tweak: reworks hyposprays into the Apollo.
balance: Swaps out bicard from hypokits to polystem in the renamed MID kits.
balance: buffs polystem to heal faster, rather then being objectively worse then tricord, and far harder to make.
imageadd: added new MID sprites.
```


Jamini (approved):
Right, we've discussed it to death. Looks  gucci

PR #367, opened 2021-04-04
Lore update and a map change

Ikmachek:
## About The Pull Request
So this PR does two things. Firstly it's another installment in changing object names to fit lore, and I TRIED to comment out the wiki manuals that are underappreciated and we can never used but it tossed compile errors so I just replaced the engine and forensics guide with little 'how do to do basics' papers.

## Why It's Good For The Game
I like the charm of instead of a technical manual that doesn't even work to explain the engine, it's some engineer's scribbles for others to find. This SHOULD be modular. However it wasn't because I was commenting out stuff and given that didn't work, I was too damn lazy to make the paper scrap objects all over in a modular folder, recompile them, remap them, etc. Later I might even find a way to get rid of those old, useless wiki manuals.

## Changelog
```changelog
add: Two new engine and forensics guides in paper scrap form!
tweak: Changes some object names to un-eris them.
```


zzz_modular_syzygy/code/modules/loadout/lists/uniforms.dm
```diff
@@ -24,3 +24,7 @@
 	display_name = "black turtleneck"
 	path = /obj/item/clothing/under/syndicate/civilian
 
+/datum/gear/uniform/tacticalturtleneck
```
zackerycodestoo, line 27: Non-blocking change; I would suggest moving this to its own file in `zzzz_modular_occulus` rather than editing `zzz_modular_syzygy`

zackerycodestoo (approved):
Approved; only a minor non-blocking suggestion in comments

zzz_modular_syzygy/code/modules/loadout/lists/uniforms.dm
Jamini: the syzygy folder is technically one of our modular folders anyway. So this is fine. It'll end up getting moved eventually when I get to it as I'm slowly moving those files to our folder anyway.

PR #368, opened 2021-04-04
postmortum restorative patch

Jamini:
## About The Pull Request

Adds in surgeries and chems that can lower brute and burn damage on patients just outside of defibable range

image

## Why It's Good For The Game

This makes the -100 damage cap a bit less arbitrary.
These options are not for someone with excessive damage. Past a certain point you should be resleeving or regenerating someone.

## Changelog
```changelog
add:  burn treatment surgery option
add: trauma treatment surgery option
add: reconstructive nanites
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/surgery/organic_damage.dm
+++ b/zzzz_modular_occulus/code/modules/surgery/organic_damage.dm
@@ -1,0 +1,63 @@
+/datum/surgery_step/fix_brute
+	allowed_tools = list(
+		/obj/item/stack/medical/advanced/bruise_pack = 100,
+		/obj/item/stack/medical/bruise_pack = 20,
+	)
+
+	duration = 80
+
+/datum/surgery_step/fix_brute/can_use(mob/living/user, obj/item/organ/external/organ, obj/item/stack/tool)
+	return BP_IS_ORGANIC(organ) && organ.is_open() && organ.brute_dam >= 15
+
+/datum/surgery_step/fix_brute/begin_step(mob/living/user, obj/item/organ/external/organ, obj/item/stack/tool)
+	user.visible_message(
+		SPAN_NOTICE("[user] begins to mend [organ.get_surgery_name()]."),
+		SPAN_NOTICE("You begin to mend [organ.get_surgery_name()].")
+	)
+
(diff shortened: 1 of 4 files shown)


zzzz_modular_occulus/code/modules/surgery/organic_damage.dm
```diff
@@ -0,0 +1,63 @@
+/datum/surgery_step/fix_brute
+	allowed_tools = list(
+		/obj/item/stack/medical/advanced/bruise_pack = 100,
+		/obj/item/stack/medical/bruise_pack = 20,
```
zackerycodestoo, line 4: Non-blocking; what does the number for the tool mean? Is that an analogue for the tool quality?

zackerycodestoo (dismissed):
I have two concerns:

1. Would you be able to add feedback when clicking the 'Mend' button without the medicine in hand so the doctor knows what they need? Logically they could come to the conclusion of a bruise pack, etc. but as of right now, nothing happens when you click 'Mend' and you don't have the right medicine in hand.
2. The success chance for regular bruise packs _might_ be too low -- I had 60 BIO and I was using a regular bruise pack but I was constantly damaging the patient anyway. I know it's intended to use the advanced packs, but there's nothing to indicate that (beyond code diving) currently.

zzzz_modular_occulus/code/modules/surgery/organic_damage.dm
Jamini: Yes. As bruise packs don't have that quality.

Jamini:
> I have two concerns:
> 
>     1. Would you be able to add feedback when clicking the 'Mend' button without the medicine in hand so the doctor knows what they need? Logically they could come to the conclusion of a bruise pack, etc. but as of right now, nothing happens when you click 'Mend' and you don't have the right medicine in hand.
> 
>     2. The success chance for regular bruise packs _might_ be too low -- I had 60 BIO and I was using a regular bruise pack but I was constantly damaging the patient anyway. I know it's intended to use the advanced packs, but there's nothing to indicate that (beyond code diving) currently.

1. I followed the same code we have for other surgeries. I will double check if I missed anything. There is a possibility to add feedback.
2. The success chance for a regular bruise pack is the same as organ repair. I'd rather keep that consistent.

zackerycodestoo (approved):
Great!

PR #366, opened 2021-04-04
Text "Volume" now respects selected font size.

gozulio:
## About The Pull Request
fixes #364 
Volume scaling for says, witnessed examines, and pain messages will now respect the users selected font size.
The messages originally used html font size selectors instead of css span selectors. This PR changes them to use css selectors and em units for font size.

## Why It's Good For The Game
Improve accessibility for the visually impaired.

## Changelog
```changelog
fix: fixed font size scaling for certain chat messages.
```

PR #372, opened 2021-04-08
Font scale patch

gozulio:
## About The Pull Request
This is a follow-up PR for PR #366
This PR fixes several things I missed in the previous PR, Specifically radio messages will no longer be tiny.

## Why It's Good For The Game
This PR Fixes multiple issues with PR #366 

## Changelog
```changelog
tweak: Reduced font size for shouting
tweak: Reduced volume affect for stims and toxins that affect speech volume.
fix: Contents of radio messages will now scale font size correctly.
fix: Tape Recorder now records speech volume in em units instead of pixels. (to match other text scaling.)
```


gozulio:
Radio messages
image

Shouting not being as loud
image

Shouting while high on violence stim.
image

Jamini (approved):
Right, we are going to try merging the fix first, and if that doesn't work we will revert both patches.'

Approved.

PR #376, opened 2021-04-09
Buffing hivemind and blob spawn chance.

Jamini:
## About The Pull Request

Hivemind player requirement reduced to 6. 9 for full weighting.

## Why It's Good For The Game

We have people who have never seen either on Occulus

## Changelog
```changelog
balance: hivemind player requirement reduced from 9 to 6. hivemind ideal playercount reduced from 12 to 9
```


zackerycodestoo:
Was there already a discussion on this? I'm fine with it spawning on Sentinel but Healer indicates it does not have round-domineering events, I would think a spreading blob would qualify for that

Jamini:
> Was there already a discussion on this? I'm fine with it spawning on Sentinel but Healer indicates it does not have round-domineering events, I would think a spreading blob would qualify for that

Healer and Sentinel both do not have round domineering events. Blobs can be relaibly soloed by one person and AFAIK, do not end the round on Eris code.

Jamini:
Updated as per emojicracy.

Michiyamenotehifunana (approved):
Simple number changes, should be good.

PR #386, opened 2021-04-14
fixes lag on returning to lobby

Jamini:
## About The Pull Request

This fixes the lag when the server fails to start, by pushing the initial asset distribution into the first loop only.

## Why It's Good For The Game

Big bug fix.

## Changelog
```changelog
fix: The server should no longer lag terrible for everyone when the game fails to start because of no players.
```


Michiyamenotehifunana (approved):
Nothing looks egregious, just some shuffling of proccall orders. Approved.

PR #404, opened 2021-04-21
(Proof of concept) Minimum identification levels for reagents

zackerycodestoo:
## About The Pull Request

**This is a proof of concept. This is not a true pull -- the intent of this pull is to show that the idea is viable without significant code alterations, and I am unlikely to personally continue work on it.**

This non-modular adds the flag `minimum_identification` to `/datum/reagent`, and changes the examine call to call reagent's new proc, `identify_reagent` rather than a direct `to_chat` with the information.

The purpose of `minimum_identification` is to act as a minimum level for identifying the liquid without a scanner.

Straight reagents, like carbon or acetone, are set to a minimum level of STAT_LEVEL_GODLIKE (100)
Medicines are set to a level of STAT_LEVEL_EXPERT (40)
Alcohol is set to a level of STAT_LEVEL_BASIC (15)
Drinks are set to a level of 0.	(space cola, dr. gibb)
Water is set to 0 (because normally it would try to inherit 100 from `/datum/reagent`)

There is also a small check to see the 'exact amount' of a liquid based on COG. It's easy to pass at a 15 minimum, otherwise you'll just get a descriptor saying how much is in the glass.

Screenshot examples (debug messages are not in live code, and descriptor texts are capitalized in live code):

Acetone (min 100)
image

Water (min 0)
image

**COG below 15:**

Absinthe (min 20)
image

Space cleaner (min 100, could probably be lowered)
image

Finally, silicons ignore the skill requirements. This does not touch reagent descriptions or names (as you can see the final picture has 'contains space cleaner'. The desc could also be obfuscated if it's an unknown reagent but that is not in the current code.

## Why It's Good For The Game

More obfuscation for reagents -- now a vagabond can't look a vial and go 'oh! that's ossisine!'

## Changelog
```changelog
balance: Skill checks are required to identify reagents and how much of it there is
```

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -2909,6 +2909,7 @@
 #include "zzzz_modular_occulus\code\modules\projectiles\projectile\bullettypes.dm"
 #include "zzzz_modular_occulus\code\modules\projectiles\projectile\projectile.dm"
 #include "zzzz_modular_occulus\code\modules\radio\radios.dm"
+#include "zzzz_modular_occulus\code\modules\reagents\chemical_identification.dm"
 #include "zzzz_modular_occulus\code\modules\reagents\food-Drinks.dm"
 #include "zzzz_modular_occulus\code\modules\reagents\recipes.dm"
 #include "zzzz_modular_occulus\code\modules\reagents\reagent_containers\hypospray.dm"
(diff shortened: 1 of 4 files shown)


Jamini:
This is well done and a much bigger and better scope than I was going to do. It looks great.

I fiddled with some values. The base identification value is 0 in a stat (there aren't many of them anyway), while various classes if reagents are harder or easier to ID.

Only Nanites were given a value of 80. Making it damn hard for anyone to know what they are.

Jamini:
Since I touched on this Tori is going to need to approve it.

Jamini (approved):
I looked it over, but I'd like Tori to double-check before merge

Michiyamenotehifunana (approved):
Nothing looks particularly egregious here.

PR #410, opened 2021-04-23
Ports 5806 - 5821

bearrrrrrrr:
## About The Pull Request

big pull request time. as per request from jams.

5806 to 5821


Jamini (approved):
Ports up to 5823

I will be handling 5809, 5811, 5820
5811 will be overwritten via modular files
5257 is a prerequisite 09 and 20

PR #433, opened 2021-05-05
Horseshoes

Ikmachek:
## About The Pull Request
PR adds a cute little set of horseshoes as a foot object and into the loadout. Tested running with and without them and they function. The icon is a little large but I'll just do what I always do when I run into problems with my code - Blame Tori.

## Why It's Good For The Game
Because while for now we only have one handsome, smart and talented equine-legged character on Occulus, I figured eventually more hooved characters will be made and this is so they don't have to wear sneakers or something just to not die on glass shards slower than a sloth's gait.

## Changelog
```changelog
add: horseshoes
```


cev_eris.dme
```diff
@@ -349,7 +349,7 @@
 #include "code\datums\music_tracks\comet_haley.dm"
 #include "code\datums\music_tracks\downtown.dm"
 #include "code\datums\music_tracks\elevator.dm"
-#include "code\datums\music_tracks\engage.dm"
+#include "code\datums\music_tracks\Engage.dm"
```
Jamini, line 352: This shouldn't be capitalized Ikky. Can  you fix the filename?

We basically should never have folders or files capitalized except in very specific circumstances.

Jamini: This is the only blocker. So once you fix that we should be gucci.

Jamini (changes requested):

Jamini (approved):

PR #452, opened 2021-05-12
Removes fluid pipes from church, fixes A thing

bearrrrrrrr:
## About The Pull Request

removes fluid pipes from church
moves hardsuits into proper position in science, makes them _science_ hardsuits instead of rnadom hardsuits
 
## Changelog
```changelog
tweak: removes fluid pipe from churches, makes hardsuits in science proper hardsuits
```


Jamini:
We've determined fluid pipes are not the source of lag. Can you please emojicracy the fluid pipes? I know I and a few others would want to keep them because pulling canisters is really ass.

ShadzFen:
PR became outdated, emojiocracy resulted in keeping the pipes. Closing PR.

PR #467, opened 2021-05-20
Adds club account access to club workers

razekpraxis:
## About The Pull Request

This change adds the club account to the club worker's memory on spawn. This allows club workers to use the EFTPOS and ATM.

## Why It's Good For The Game

As it stands, club workers are unable to do anything with the cash that the club makes beside putting it in the back room or carrying it. 

This also means that club workers are unable to use the EFTPOS machine unless:

1) there is a club manager present
2) they use their own personal account

Club already suffers from low population, and doesn't get any better for managers. Adding this will just be QoL change for workers.

## Changelog
```changelog
add: Added club worker access for club account
```

### DIFF
--- a/code/game/jobs/job/civilian.dm
+++ b/code/game/jobs/job/civilian.dm
@@ -57,6 +57,7 @@
 		STAT_TGH = 10,
 		STAT_VIG = 5,
 	)
+	department_account_access = TRUE
 	outfit_type = /decl/hierarchy/outfit/job/service/waiter
 	description = "As a Club Worker, you work for the Club Manager. Your job is to fulfill your duties in running the Club and making sure all the customers are satisfied.<br>\
 	<br>\


razekpraxis:
please be gentle

Jamini:
This was intentionally not done when we initially gave club workers a wage. The entire purpose of the Club Manager is to manage the civic account and bar staff. Giving club workers this is very much too far, especially as lorewise club workers can be literally anyone pulled off the street.

razekpraxis:
> This was intentionally not done when we initially gave club workers a wage. The entire purpose of the Club Manager is to manage the civic account and bar staff. Giving club workers this is very much too far, especially as lorewise club workers can be literally anyone pulled off the street.

if this is the case then it'd be at the very least nice to add a better way to handle or manage cash when there is no manager present.

Jamini:
Radiant found some cash registers in code. We can see about using them for taking cash payments?

razekpraxis:
> Radiant found some cash registers in code. We can see about using them for taking cash payments?

Yup. this is a pretty good solution too. adds a little bit more gameplay and is Neat.

PR #483, opened 2021-05-26
Centcom remap

Ikmachek:
## About The Pull Request
So despite Occulus almost never having any evacs, an idea came to me of what our evac centcomm should be and that idea was 'the cargo bay of the other trade ships around the NL' So I made a small, crappy little cargo bay that looks like it belongs on some rinky-dink freighter.

[Like this](https://i.gyazo.com/df05a922cfb74b3618e42ddf5f3448cd.png)

## Why It's Good For The Game
Imagine the crew has to evacuate... They barely make it out, and what awaits them is a cargo bay with some medical supplies and rollers quickly set up by another ship not expecting refugees so soon. It's grungy, people are hurt and the best they have is some medical kits, surgery tools and roller beds. Glorious.

## Changelog
```changelog
add: A new centcomm for pods to dock at
```


Jamini:
This could... use a fair amount of polish imo

ShadzFen:
Would this be connected to an actual ship? or to centcomm itself? If so that would be pretty cool. 

But if it's just a cargo bay by itself floating in space it might actually be a downgrade from just it being centcomm.

Ikmachek:
It's intended to be the cargo bay to a ship the FTU trades with like the Caudacus, or Hellcat. It's just locked so I don't have to... you know, map a whole ship.

Ikmachek:
This is about as done as I'm going to get it. I'm really not comfortable using simulated tiles then trying to have to find power for lights, etc. Even if it's not a great map, it's better than what we have now and can be fixed up later by someone who's a better mapper than me.

Ikmachek:
https://i.gyazo.com/c7add8371a0a38af4b3748cc3c783524.png 

Forgot picture

Jamini (approved):

PR #485, opened 2021-05-27
Hand Retail Scanners

RadiantFlash7:
## About The Pull Request
Adds in Retail Scanners, function the same as the new cash registers, only it takes cards instead of cash like a register. Also, portable!
Uses the same code that the registers do to identify prices, and also has the ability to create custom prices per order to alleviate any goods mispriced.

## Why It's Good For The Game
Basically allows portable registers that we put onto the map! Works almost the exact same way as the registers for consistency, and portability for convenience! 

## Changelog
```changelog
add: Added Retail Scanner
imageadd: added retail scanner Icons
imagedel: deleted some icons and images
```


Jamini:
Once again, these really don't do anything the EFTPOS already does. But that's just my thought on it.

RadiantFlash7:
> Once again, these really don't do anything the EFTPOS already does. But that's just my thought on it.

They're generally just better usage wise to be less jank and more streamlined. Easier to use and look better interface wise.

Jamini (approved):
Let's see if they are useful then.

PR #488, opened 2021-05-27
Ports shield updoots

Michiyamenotehifunana:
## About The Pull Request

Shields now have innate damage reduction (read: armor) and will render the user immune against roach knockdown (and also knockout knockdowns from melee). However, you now need to be walking to cover yourself with a shield properly. Shield effects do not apply at all unless you are doing so.

I have left the mob blocking code bits we added alone.

Also shields now apply a temporary slowdown effect to a target if you bash a living thing with it.

I had to kludge Altyn code into shields to make them actually function and not runtime. Keep that in mind in case upstream also fixes that.

## Why It's Good For The Game

Shields are now much more impactful and balanced with this. You become very slow when actively blocking, and attempting to stack these effects with strong armor would make you a snail, even if you tried to powergame all the speed boosting effects at once.

## Changelog
```changelog CEV-Eris Contributors
balance: Shields now have innate damage resistance and can defend against certain types of knockdowns. You have to be walking to properly use shields now. Shield bashing applies a slowdown effect to the target.
```


Jamini (approved):
Bugfix, or bugsource? Who knows

But it looks good!

PR #498, opened 2021-05-31
Fixes the multieye bug, makes rejuv work properly

Michiyamenotehifunana:
## About The Pull Request
After half a year, I have finally figured out how to make people stop spawning with 3 sets of eyes, one of which becomes a null reference a minute and a half after spawning, thus causing mysterious bouts of blindness.

I have also made rejuv not eject core implants (the upstream fix was not quite compatible with the way we did things), but it may occasionally still vomit out a duplicate implant for reasons unknown. Rejuv also does not wipe body markings anymore.

## Why It's Good For The Game

bugfixes good
bugs bad

## Changelog
```changelog Toriate
fix: players should no longer spawn with 3 sets of buggy eyes
fix: rejuv should work without ruining characters now
```

### DIFF
--- a/code/datums/setup_option/core_implants.dm
+++ b/code/datums/setup_option/core_implants.dm
@@ -10,15 +10,14 @@
 	allow_modifications = TRUE
 
 /datum/category_item/setup_option/core_implant/cruciform/apply(mob/living/carbon/human/character)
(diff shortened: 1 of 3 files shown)


Jamini (approved):
Does what it says in a tin.

PR #503, opened 2021-06-03
Fixes duplicate core implants and miscellaneous runtimes

Michiyamenotehifunana:
## About The Pull Request

Now the related procs check if you have an existing core implant before trying to install a second one. Also added checks that should prevent pointless code from running on mannequins and runtiming.

## Why It's Good For The Game

bugfix

## Changelog
```changelog Toriate
fix: players should no longer spawn with duplicate core implants
```


Jamini (approved):
Looks solid.

PR #504, opened 2021-06-03
Rubberizes Rubber Bullets

Zenitheevee:
## About The Pull Request

Removes rubber bullet's ability to overpen targets on .20, .25 caseless and .30 (The only rubber rounds that overpens). They're rubbers, not HV/Lethal rounds

I've tested and this seems to fix the overpenning while being modular and without fucking gun damage values

## Why It's Good For The Game

I hate accidentally shooting friends because i'm using rubbers that aren't rubbery enough

## Changelog
```changelog
balance: Gives caseless 25, .20 and .30 rubber rounds a "Penetrating = 0" variable
```


Jamini (approved):
A simple var override (RIP overenning roaches with my sol)

PR #505, opened 2021-06-04
Changed the alert level wording

Ikmachek:
## About The Pull Request
This NON-MODULAR CHANGE edits JUST THE WORDING of the alert level ups and downs to more accurately reflect our lore and setting, I.E. we're not a station and no one cares security has a baton on display.

Why is this not modular? They're procs, that's why.

## Why It's Good For The Game
Back in the day these alert messages were aimed for Bay and extremely security-centrist. More aimed around the SOP that security couldn't even have weapons visible on code green and such. that was 8 years ago and this is Occulus where everyone has guns out and there's more threats to the ship than Aegis' wheelhouse.

## Changelog
```changelog
fix: Fixes the wording in alert levels to reflect the Northern Light more
```

### DIFF
--- a/code/datums/security_state.dm
+++ b/code/datums/security_state.dm
@@ -192,7 +192,7 @@
 		if(FA.z in GLOB.maps_data.contact_levels)
 			FA.update_icon()
 	post_status("alert")
-
+/////Occulus Note: Changing the wording of the alert levels, nothing else/////
 /decl/security_level/default/code_green
 	name = "code green"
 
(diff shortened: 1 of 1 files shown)


Jamini (changes requested):
/decl/security_level/default/code_blue is not a proc, you should be able to override this., same with the others.

Ikmachek:
Okay because I guess a /delc function can be modular and we're messing with the bot I'm going to close and rework this.

PR #508, opened 2021-06-06
Deletes all roundstart monkeys, and more

Michiyamenotehifunana:
## About The Pull Request
Monkeys are basically never used, and hog up what little processing power we have. Monkey cubes take their place.
StrongDMM_J2dijFpBAx

Also since this is a mapping thing I have taken the liberty of adding a tank dispenser to the bridge so that the space suits there actually have utility.
StrongDMM_02EtIM1xME

## Why It's Good For The Game
Less lag, less log spam. Also the bridge space suits are no longer just decoration.

## Changelog
```changelog Toriate
add: The bridge now has a tank dispenser for the space suits present there.
del: monki
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -20556,9 +20556,9 @@
 /turf/simulated/wall/r_wall,
 /area/eris/maintenance/substation/section1)
 "aXi" = (
-/mob/living/carbon/human/monkey,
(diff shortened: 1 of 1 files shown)


Jamini (approved):
I agree with this change, but this is a bandaid fix.

The duplicate soulcrypt issue and crypts/Cruciforms spawning on mannequins is still a serious issue and needs to actually be addressed  please.

PR #509, opened 2021-06-06
Fixes the SC updater eating IDs, fixes holochip init

Michiyamenotehifunana:
## About The Pull Request
The Soulcrypt Access Updater would delete an inserted ID if you tried to insert a new ID in without taking out the old one. It will now prevent that and notify the user that there is already a card within.

Holochips have also had their init proc callback to the parent, which should fix the atom initialization oddities we've been experiencing. Keyword is should. It did not fix it on my test server.

## Why It's Good For The Game

Less bugs!

## Changelog
```changelog
fix: SC Access Updater should no longer delete inserted IDs when you insert another one in
```


Jamini (approved):
Huh, I'm glad  we caught this.

PR #507, opened 2021-06-06
Fixes duplicate core implants without rebreaking everything

Michiyamenotehifunana:
## About The Pull Request

Now all redundant procs have been moved into where they ought to be, and useless checks have been removed. This fix has been tested and confirmed to work.

## Why It's Good For The Game
so you guys can stop hounding my ass about this

## Changelog
```changelog
fix: players no longer spawn with duplicate core implants.
```

PR #510, opened 2021-06-06
Revert "Fixes duplicate core implants without rebreaking everything"

Jamini:
Reverts Occulus-Server/Occulus-Eris#507


ShadzFen (approved):
Test for lag fix.

PR #512, opened 2021-06-06
Revert "Fixes the multieye bug, makes rejuv work properly"

Jamini:
Reverts Occulus-Server/Occulus-Eris#498


ShadzFen (approved):
Test for lag fix.

PR #518, opened 2021-06-07
Revert "Revert "Fixes duplicate core implants without rebreaking everything""

Jamini:
Reverts Occulus-Server/Occulus-Eris#510

The last revert of our revert


Michiyamenotehifunana (approved):
revert a reverted fix for a fix for a fix

PR #519, opened 2021-06-07
Handtele now works on all valid zlevels

Michiyamenotehifunana:
## About The Pull Request

Exactly what it says on the tin. Handtele was preventing teleports from Zlevel 2 because """""it does not exist""""". That restriction has been removed.

## Why It's Good For The Game

less jank

## Changelog
```changelog Toriate
fix: handtele should now work on all valid zlevels instead of malfunctioning on deck 4
```


Jamini (approved):

Jamini:
This is old code. The "Does not exist" refers to levels above 7. Originally z-level 2 was CC

PR #527, opened 2021-06-08
Sets feet blood DNA to null upon cleaning.

fludd12:
## About The Pull Request
As far as my understanding goes, overlays are not cut unless feet_blood_DNA is null, and cut sets it to an empty list.

## Why It's Good For The Game

This should resolve the overlay not being properly removed when washing feet.

## Changelog
```changelog
fix: wosh u feet
```


Michiyamenotehifunana (approved):
confirmed to be working, good to go

RadiantFlash7:
Is this just feet by themselves, or boots? I know currently boots won't update their sprite when cleaned if they have a boot knife within.

PR #528, opened 2021-06-09
Extends GC hard delete warning time to 2 seconds

Michiyamenotehifunana:
## About The Pull Request

Currently, the server seems to be having trouble GC'ing a lot of things, taking over 1.3 seconds to delete roaches and such. When this happens, the subsystem intentionally hangs for the exact same about of time that it took to delete for reasons that go over my head. This causes the server to stutter noticeably. I think.

I am unable to test this as my local hardware does not have trouble GC'ing anything. Either way, it compiles and runs just fine.

## Why It's Good For The Game

Anything to make the server more playable is preferable.

## Changelog
```changelog Toriate
server: GC should now only error when something takes longer than 2 seconds to delete
```

### DIFF
--- a/code/controllers/subsystems/garbage.dm
+++ b/code/controllers/subsystems/garbage.dm
@@ -230,9 +230,11 @@ SUBSYSTEM_DEF(garbage)
 		time = TICK_DELTA_TO_MS(tick)/100
 	if (time > highest_del_time)
 		highest_del_time = time
-	if (time > 10)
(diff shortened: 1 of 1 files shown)


Jamini (approved):
Worth looking at. This is definitely a hardware issue, possibly related to the fact that we are on a quad-core processor (and thus, have less processor power as byond only uses one core) now.

I'll look into a solution with our new host as well.

PR #531, opened 2021-06-10
A Surprisingly Painless Port of Erismed 3

Michiyamenotehifunana:
## About The Pull Request
Pain is now overhauled. Refer to https://github.com/discordia-space/CEV-Eris/pull/5915

I have reduced the paincrit knockout time from 90 seconds, to a range of 15-25 seconds. I believe we should tweak this further to depend more on stats. It should also be noted that most pain is removed upon waking up from a knockout, so this may present another dimension in balancing.

I have also taken the liberty of porting (and actually rendering functional) the tweaks that make non-vital limbs (e.g. not chest/head) not contribute to actual health. This makes it possible to lose several limbs and still keep on trucking on. It is also possible to safely put someone into paincrit by lethaling their non-vital bits now.

More augmentation options are available now (specifically internal organs). We are missing the full FBP update that adds Eris branded FBP parts.
mclotS7me1

This has been compiled, tested, confirmed to be working, and so far does not appear to runtime or have egregious bugs.

## Why It's Good For The Game

Is this truly good for us? I don't know, only time will tell. A lot of challenges stand before us in balancing this for a fun RP environment, but it certainly will be exciting.

## Changelog
```changelog Toriate
add: Erismed3. Pain is completely overhauled. Paincrit knocks you out, non-vital organs don't count torwards overall health.
```


Jamini:
Can you check a few things for me Tori?

1. Xenobiology. Moostard reported that slimes are severely bugged and don't/cant kill their targets. can we confirm cloneloss can kill?
2. Bleedout rates. Is our bleeding high enough to cause people to die from bloodloss in a reasonable timeframe?

Michiyamenotehifunana:
1. Slimes do not deal cloneloss. They only deal burn and toxin damage. Having a slime latched onto you for a minute or so will put you into paincrit, and after a few more moments, the slimes will automatically detach as they will only latch on up to a certain health threshold (150 burn damage). Any inability to kill seems intentional.
2. Yes. Very much yes. You can roll around as a funny nugget man if all your limbs get pulped, but not for long. Even if the limbs are robotic.

Jamini:
image

False. Slimes deal cloneloss.

Jamini:
>     1. Any inability to kill seems intentional.

Eris values re not our values. I'll approve this, but keep in mind that I'm going to be bringing back highly lethal Xenobio soon.

Jamini (changes requested):
Our versions of Bicard and Merealyn need their pain-causing values changed to match paracetamol/Tramadol.

Jamini (approved):

PR #556, opened 2021-06-16
Elevator to the dorms!

ShadzFen:
## About The Pull Request
This PR adds an elevator to the dorms area that allows players to leave the current round early by returning to the dormitories. 

The elevator will remove any player or living mob from the shift and will announce them leaving similarly to the cryostorage units. Any living mob within the elevator area will be removed shortly after the elevator button is pressed.

**A clear warning is given to all living beings that enter the elevator area that they are at risk of being removed from the shift if someone else in the elevator presses the elevator button.** 

image
image

Nerd stuff and known bugs:
- The elevator area is using infinite power. It does not rely on the station's power. This is because, as a location for exiting the shift, it is considered a critical area.
- The elevator area is given the critical area flag.
- The elevator area plays its own ambience music, which causes it to play after being ghostized, even when leaving the area. Unsure how to fix this issue.
- Preserved items inside a container that is inside a container (e.g., a spare captain's ID that is inside a box, that is inside a bag) will get deleted and not dropped on the ground like the rest of the preserved items. Unsure how to fix this for now.
- Preserved items inside a jacket or coat pocket will also get deleted rather than dropped. Unsure how to fix this.
- Upon entering the elevator area there is a considerable (roughly half a second) stutter. This is due to it loading the elevator_music.ogg file. It is roughly similar to the stutter you get when first entering maintenance on a shift. Once loaded a single time, it will not stutter again. Given the circumstances of the elevator being a round exit method, I believe this stutter is insignificant, however it's a bug nontheless. I do not know how to fix it without severely reducing the size of the .ogg file.

Commit 1 post PR creation:
This commit fixes the following:
- The elevator call button can no longer be spammed. It will wait for the elevator to finish moving players to the dorms before being able to be called.
- The warning prompt will now display correctly, asking players that click the elevator panel whether they are sure they want to leave the round or not. 
- Some code optimization is implemented to avoid checking every single airlock and every single machine. Now it just checks within a range of 5 tiles.
- The elevator will now fail to proceed if the airlock can not be closed and bolted, displaying an IC error message.
- Some of the code has been commented to help explain the entire processs of the elevator's functions.
- Preserved items will now properly drop, rather than get deleted. 

Commit 2 post PR creation:
This commit fixes the following:
- Some small code issues with `&` not being `&&`.

Commit 3 post PR creation:
This commit fixes the following:
- The elevator area is now a rad storm protected area. 
- BRAND NEW ELEVATOR PANEL SPRITES courtesy of Toriate himself. Go and say thank you to him.
- Elevator panel displays itself going up the floors of the ship all the way to floor 6, a fictional floor that is fluidly represented in lore, but not in game. This floor consists of dorm rooms for the crew.
- More robust checks are in place to completely prevent the elevator from working when someone presses the button while standing in the airlock.
- A dorms_elevator.dmi file now exists. An image will be attached in a comment below.
- An error sound plays when the elevator malfunctions along with the error message.
- Elevator has been given a small_softfloor sound environment flag in order to properly simulate what other sounds would be like inside the elevator.
- Code has been further commented for better understanding.

## Why It's Good For The Game
As a HRP server, we should strive for additions that increase the immersion of the game. This is a feature many have wanted for quite some time now, and while it may still be in a basic stage, it will be more complete once I get the fix PR out along with the small details and features.

## Changelog
```changelog
add: Elevator in the dorms area that gives players an alternate way of leaving the round early.
add: Elevator area in the map code, as well as dedicated elevator buttons.
soundadd: Elevator audio / music.
```


zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+	update_icon()
+	..()
+
+/obj/machinery/button/remote/elevator_panel/trigger()
+	var/elevator_prompt
+	var/turf/T = get_turf(src)
+
+	if(world.time - cooldown_timer > cooldown)
+		cooldown_timer = world.time
+		elevator_prompt = alert(src, "Warning! Using the elevator will end your shift and remove you from the round. Are you -sure- you want to head to your dorm for the day?\n(Everyone inside the elevator with you will end their shift and be removed as well!)", "End your shift and head to the dorms?", "End shift", "Stay on shift")
```
Shadow-Quill, line 75: You're attempting to show the alert to the thing calling the proc, which (if I'm not mistaken), is the panel.
I think using usr instead of src would work here.

ShadzFen:
Github isn't letting me edit my PR so... I'll just post this picture I forgot to add as a comment, showing the warning and the departure announcement.

image

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+					D.lock()
+			sleep(10)	//Wait 1 seconds before proceeding
+			playsound(src, 'zzzz_modular_occulus/sound/effects/elevator_going_up.ogg', 70, 0)
+			sleep(10)
+			visible_message("<span class='notice'>The elevator hums and creaks as it makes its way to the dormitory floor.</span>")
+			sleep(185)	//Wait 6 seconds before proceeding
+			for (var/mob/living/mobsgoingup in get_area(src))
+				despawn_passenger(mobsgoingup)
+		else
+			T.audible_message("<font color = #F63812><B>Elevator Coms</B>: \"Error. The elevator is currently out of service. Please contact your local engineering department.\"</font>")
```
Shadow-Quill, line 90: This would announce the error if the person interacting with the button decides to stay on shift.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+	if(world.time - cooldown_timer > cooldown)
+		cooldown_timer = world.time
+		elevator_prompt = alert(src, "Warning! Using the elevator will end your shift and remove you from the round. Are you -sure- you want to head to your dorm for the day?\n(Everyone inside the elevator with you will end their shift and be removed as well!)", "End your shift and head to the dorms?", "End shift", "Stay on shift")
+		if(elevator_prompt == "End shift")
+			for(var/obj/machinery/door/airlock/D in SSmachines.machinery)
+				if (D.id_tag == id)
+					D.close()
+					sleep(11)
+					D.lock()
+			sleep(10)	//Wait 1 seconds before proceeding
```
Shadow-Quill, line 82: You can check if an airlock has successfully locked with:
`if(D.command_completed("secure_close"))`
from
https://github.com/Occulus-Server/Occulus-Eris/blob/6cba1431fe27a9a96fc6f9e84ee09e38a741e057/code/game/machinery/doors/airlock_control.dm#L69

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+
+/obj/machinery/button/remote/elevator_panel/trigger()
+	var/elevator_prompt
+	var/turf/T = get_turf(src)
+
+	if(world.time - cooldown_timer > cooldown)
+		cooldown_timer = world.time
+		elevator_prompt = alert(src, "Warning! Using the elevator will end your shift and remove you from the round. Are you -sure- you want to head to your dorm for the day?\n(Everyone inside the elevator with you will end their shift and be removed as well!)", "End your shift and head to the dorms?", "End shift", "Stay on shift")
+		if(elevator_prompt == "End shift")
+			for(var/obj/machinery/door/airlock/D in SSmachines.machinery)
```
Shadow-Quill, line 77: Maybe using an `in range(src, <somenum>)` (instead of `in SSmachines.machinery`) would be better here, instead of having to loop through every single door. Since (I presume) the elevator button will always be in close proximity to the airlock.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+				if (D.id_tag == id)
+					D.close()
+					sleep(11)
+					D.lock()
```
Shadow-Quill, line 81: You can use `D.do_command("secure_close")` to do this in one line.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+/obj/machinery/button/remote/elevator_call_button
+	name = "Elevator Call Button"
+	desc = "An button used to call the elevator down to the current floor."
+	var/cooldown = 24 SECONDS
+	var/cooldown_timer
+
+/obj/machinery/button/remote/elevator_call_button/trigger()
+	for(var/obj/machinery/door/airlock/D in SSmachines.machinery)
+		if (D.id_tag == id)
+			if(D.locked)
```
Shadow-Quill, line 103:
```suggestion
		if(D.id_tag == id && D.locked)
```

Shadow-Quill: If you change this, 104-117 will need to be un-indented.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+				if(world.time - cooldown_timer > cooldown)
+					cooldown_timer = world.time
+					visible_message("<span class='notice'>The elevator has been called. Please wait...</span>")
+					playsound(src, 'zzzz_modular_occulus/sound/effects/elevator_going_down.ogg', 70, 0)
+					sleep(220)
+					for(D in SSmachines.machinery)
+						if (D.id_tag == id)
+							D.unlock()
+							sleep(10)
+							D.open()
```
Shadow-Quill, line 113: You can use `D.do_command("secure_open")` to open and bolt the airlock open, though if you just want to have it unbolted (and not bolted open), this should be fine.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
+///// Dorms Elevator/////
+
+//// This code is for the elevator to remove the player from the round and delete all items outside of the preserve items list.
+
+/obj/machinery/button/remote/elevator_panel/proc/despawn_passenger(var/mob/living/user)
+	var/mob/living/passenger = user
+
+	for(var/obj/item/W in passenger)
+		if(W in preserve_items)
+			passenger.drop_from_inventory(W)
```
Shadow-Quill, line 10: Is there a way to link this to a cryo console, so that special items aren't just there for the taking, or does this do that automagically?

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -0,0 +1,131 @@
...
+			else
+				D.open()
+
+
+//// This is the code for the elevator area.
+
+/area/eris/elevator
+	name = "elevator"
+	forced_ambience = list('zzzz_modular_occulus/sound/ambience/elevatormusic.ogg')
+	requires_power = 0
```
Shadow-Quill, line 125: FALSE

Shadow-Quill: (Though if this isn't the standard practice for on/off variable, feel free to ignore.)

Shadow-Quill (changes requested):

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
Shadow-Quill: Alternatively, just grab the area and check airlocks in that.

Shadow-Quill:
As for the elevator call button only checking if the airlocks are locked, throw a variable on something (prolly the elevator internal button) that keeps track of if the elevator is in transit, then have the call button check that.

ShadzFen:
Okay... damn that's a lot of good feedback.

Shadow-Quill:
> 
> 
> Okay... damn that's a lot of good feedback.

I aim to please C:

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: This is just a placeholder for now. Since the proc wasn't working to begin with, there was no way to say no, so I didn't bother moving it. It'll get changed once I get the PR fix out, along with all the other stuff as well

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Got the proc working on my branch. Not committed yet.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Got this done too, not committed yet.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: The in range one was a really good idea. A lot less intensive. Got that in place too. Not committed yet.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: That is a good point, however I changed a few things. I needed an extra check to make sure the door closed successfully before it actually locked, in order to check whether the door was malfunctioning or not. I got that sorted, not committed yet.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Cood code optimization suggestion. I always forget I can do this with if statements. Thanks! Got this on my branch now, not committed yet.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Yeah the idea is for the airlock to act like real elevators. So once it reaches the floor it will unbolt and open, then after some time it closes. Clicking the button outside the elevator while it is unbolted will open the airlock, like normal elevator buttons do.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Most of the other stuff I saw regarding areas show `requires_power = 0`

I would not be surprised if this was just Eris or SS13 code in general being bad practice, cus I have also seen many instances of `TRUE` `FALSE` being used as well. Mixed code like that is bad etiquette, but at this point I'm unsure which one is supposed to be used.  Regardless, it works so... /shrug

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: The issue with items is that the cryostorage units have their own little pockets for storing special items. I can't really do that for an elevator, since rather than being a single tile pod, they're an area on the map. I'm not really sure what to do about special items dropping. It's better for them to drop on the floor than just get deleted, cus someone could delete the captain's spare which is... bad. Speaking of...

Another bug I forgot to mention was that the list of preserve items was just getting ignored completely. The only thing that never gets deleted is the ID card inside the player's PDA. Everything else is deleted successfully, regardless of whether it's in the preserve list or not. So I need to do some testing to see if I can fix that issue as well.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
Michiyamenotehifunana: TRUE/FALSE should always be used for booleans no matter what. Eriscode is just in a sad state.

ShadzFen: Alright, I'll change it to TRUE/FALSE

ShadzFen:
PR details have been updated with the new changes from that commit. Elevator is now pretty much in fully functioning condition.

RadiantFlash7:
> PR details have been updated with the new changes from that commit. Elevator is now pretty much in fully functioning condition.

Nice job elevating the status of this PR :3

Shadow-Quill:
Mapping needs deconflicting, and I'll look over the code again in a minute.

ShadzFen:
I think the bug regarding the audio not returning to ship ambience rather than elevator ambience is because the crew quarters area doesn't have an ambience track set at all. 

Would adding the standard ship ambience ogg to this area fix that issue?

https://github.com/Occulus-Server/Occulus-Eris/blob/abd349f697e42346b096c9714a1cb210eb7de326/maps/CEVEris/_Eris_areas.dm#L380

Shadow-Quill (changes requested):

ShadzFen:
A new DMI file has been created! Here are the sprites for the elevator's panel. These will replace the button sprite INSIDE the elevator. Not the call button outside the elevator.

image

ShadzFen:
Map conflict resolved. Elevator is ready for server use. Here's a picture of the final product. 
I do still have a few ideas in mind for it, but they may take a while longer. So for now, the actual thing works and can be used without any major bugs.

image

zzzz_modular_occulus/code/game/gamemodes/events/weather/rad_storm.dm
```diff
@@ -0,0 +1,7 @@
+////Radiation storms occur when the station passes through an irradiated area, and irradiate anyone not standing in protected areas (maintenance, emergency storage, etc.)
+///This is the rad_storm modular file for Occulus. Please add any additions to the radiation storm event here. Thanks!
+
+/datum/weather/rad_storm/New()
```
Jamini, line 4: This is not the correct way to do a modular override.

Please just re-define protected_areas. Do not overwrite the new() proc

Jamini (changes requested):
Please don't use new() for modular overrides.

zzzz_modular_occulus/code/game/gamemodes/events/weather/rad_storm.dm
ShadzFen: Fixed!

Jamini (approved):

PR #559, opened 2021-06-18
Fixes adjusting clothing not updating the sprite

Shadow-Quill:
## About The Pull Request

Fixes the "Adjust clothing" verb from not actually updating the sprite.
As a note, update_icon apparently only refreshes the icon overlays, **not** updates them to changed stuff.
Go take a look at https://github.com/Occulus-Server/Occulus-Eris/blob/master/code/modules/mob/living/carbon/human/update_icons.dm to know more.

For those astute, this is a webedit. It should be fine™.

## Why It's Good For The Game

No more taking off/putting on outfits to readjust the spite.

## Changelog
```changelog
fix: Fixes the "Toggle Jumpsuit" verb. Now it correctly updates the suit to the adjusted sprite, instead of when you take off/put it back on.
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/clothing/clothing.dm
+++ b/zzzz_modular_occulus/code/modules/clothing/clothing.dm
@@ -9,4 +9,4 @@
 	usr.visible_message("[usr] adjusts their jumpsuit.", \
 	"You adjust your jumpsuit.")
 	rolldown = !rolldown
-	usr.update_icons()
+	usr.update_inv_w_uniform()


Michiyamenotehifunana (approved):
looks about right

Jamini:
Hrm? When I tested this update_icons was what worked, and update_w_uniform did *not* function in that way.

Jamini:
Weird, I could have sworn I had this working right. oh well. :shrug:

Michiyamenotehifunana:
Actually, I have my doubts that this has been tested enough for us to be absolutely sure.

Jamini:
I just hopped onto a server and confirmed this fix is correct.

Shadow-Quill:
> Actually, I have my doubts that this has been tested enough for us to be absolutely sure.

I did point out this was a webedit.

PR #561, opened 2021-06-18
Core Implant Polishing Pass and Bonus Content

Michiyamenotehifunana:
## About The Pull Request

This PR refactors Soulcrypt processing, which should make them a tad bit more optimized. In addition, it ports a litany we missed, and adds a new tab to the ritual books for our snowflake litanies so its less messy. Also it adds a cool new dagger litany.
dreamseeker_Drv17rd0N6

I originally planned to make a huge update with a lot more soulcrypt content, but this is dragging on. Depending on how soon this gets merged, I'll finish off the soulcrypt content expansion in a future PR.

## Why It's Good For The Game

Polishing, bugfixing, optimization, and new content all in one package. What's not to love?

## Changelog
```changelog Toriate
add: Immolate Litany, a crusade litany that turns a Mekhanite dagger into an unstable weapon that deals good burn and can flashbang people at range. Limited uses
add: Mekhanite energy dagger. A 10 hit dagger that deals decent burn and functions as a mini-flashbang when thrown. Harms the user to attack with.
add: Absolution Litany, a litany t hat can stabilize the target
balance: Relief buffed to 50 painkiller effect, from 20. Cost reduced to 40 energy from 50
balance: Halved Enkindle nutrition drain, removed toxin loss from Enkindle
tweak: Added a utility category for litanies.
tweak: Moved Entreaty to Utility tab
tweak: Moved Reveal to Utility tab
tweak: Moved Cruciform Sense to Utility tab
tweak: Moved Enkindle to Utility tab
fix: fixed Soulcrypts being weird.
```

PR #563, opened 2021-06-18
Fixes nutrition costs on soulcrypts

### LINKED ISSUE #475: Aegis Survival Soul Crypt program doesn't function as intended
#### Description of issue
The soulcrypt program actually adds nutrition to you, rather then removing it during use. Meaning it can be used infinitely.

Theoretically, it should drain nutrition while active, since it gives a stats buff. Instead it gives both a stat buff and adds nutrition.

#### Steps to reproduce 
Have half nutrition, 
Activate survival crypt program
Watch as it fills to green over time.

Probably just needs values adjusted in the crypt program code.

Couple days. Mentioned it to Tori a bit back, but we probably should have an issue posted so it doesn't get forgotten.

#### Issue bingo

Jamini:
## About The Pull Request

Soulcrypt active programs will now drain nutrition properly

## Why It's Good For The Game

Should fix #475

## Changelog
```changelog
fix: Soulcrypt programs now drain nutrition instead of adding to it.
```

### DIFF
--- a/code/modules/soulcrypt/base_implant.dm
+++ b/code/modules/soulcrypt/base_implant.dm
@@ -241,14 +241,14 @@ The module base code is held in module.dm
 
 	switch(nutrition_usage_setting) //Get how much nutrition we use per tick.
 		if(NUTRITION_USAGE_LOW)
-			nutrition_to_remove = 1
+			nutrition_to_remove = -1
 		if(NUTRITION_USAGE_MEDIUM)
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana:
This is already done in #561, except that it only adds a negative on the final adjustNutrition bit (i.e. adjustNutrition(-nutritiontoremove). As such, only merge one or the other, not both.

PR #567, opened 2021-06-19
Fixed the ID and implants from dropping on the floor when leaving the round via the elevator.

ShadzFen:
## About The Pull Request

IDs will now get ejected before despawning, which makes them get deleted properly. Implants will get uninstalled before despawning, which will make them get deleted properly.

The only downside to this is that the ID will make an eject sound before despawning. I might see about working on a better way of deleting the ID in the future.

## Why It's Good For The Game
Implants and IDs being dropped was pretty bad. This is a solution to that.

## Changelog
```changelog
fix: Cruciforms, soulcrypts and IDs will no longer drop on the floor when leaving the round via the elevator.
```

### DIFF
--- a/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
+++ b/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
@@ -4,9 +4,17 @@
 
 /obj/machinery/button/remote/elevator_panel/proc/despawn_passenger(var/mob/living/user)
 	var/mob/living/passenger = user
+	//var/mob/living/carbon/human/passengerplayer = user
 
 	for(var/obj/item/W in passenger)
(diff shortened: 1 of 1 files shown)


zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -4,9 +4,17 @@
 
 /obj/machinery/button/remote/elevator_panel/proc/despawn_passenger(var/mob/living/user)
 	var/mob/living/passenger = user
+	//var/mob/living/carbon/human/passengerplayer = user
```
ShadzFen, line 7: Ignore this. I was testing another method that didn't quite work and forgot to delete it. I'll delete it and info comment all this code properly once I get around to adding the next update for the elevator

Jamini (approved):

PR #587, opened 2021-07-03
Probrably stops the atuodoc from charging for autoscans

### LINKED ISSUE #351: Autodoc autocharges for scan after processes
#### Description of issue

The Autodoc, After doing any process, Will proceed to automatically draw 200 credits for another scan.

Expected behavior: Autodoc does processes and immediately stops

Actual behavior: Autodoc does processes and immediately charges 200 for another scan

#### Steps to reproduce 

Step 1: Use autodoc
Step 2: Experience extraneous healthcare costs

If i had to guess, Probably autodoc code and maybe processing code?

If i had to guess, Ever since the autodoc was installed

Sunday Highpop 21 March, We dont really have visible round ID's for this.

#### Issue bingo

Shadow-Quill:
## About The Pull Request

Adds a variable to the proc that checks if people can pay, which, if passed TRUE, doesn't try to charge the patient.
Helps stop peoples accounts from being drained just for using the autodoc (more than it would have already, that is.)

If this works, this Fixes #351 

## Why It's Good For The Game

Hidden fees are bad.

## Changelog
```changelog
fix: Fixed the autodoc charging for a rescan after each operation is done.
```

### DIFF
--- a/code/modules/surgery/autodoc.dm
+++ b/code/modules/surgery/autodoc.dm
@@ -197,7 +197,7 @@
 	if(current_step > picked_patchnotes.len)
 		stop()
 		to_chat(patient, SPAN_NOTICE("Operations complete."))	// OCCULUS EDIT: Tell the patient when it's over
-		scan_user(patient, TRUE)
+		scan_user(patient, TRUE)				// OCCULUS EDIT - Autodoc won't charge for automatically rescanning during operation
 	else
 
 		var/datum/autodoc_patchnote/next_note = picked_patchnotes[current_step]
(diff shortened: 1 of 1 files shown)


Shadow-Quill:
this just in, I remember draft PRs are a thing.

Shadow-Quill:
Finally got around to testing this, it seems that it works.

Jamini (changes requested):
Please mark changes to core files with the following comment at the end

//Occulus Edit

This helps us keep track of our changes if upstream modifies anything.

Jamini (approved):

PR #593, opened 2021-07-07
New audio system for vests and clothing in general!

ShadzFen:
## About The Pull Request

Okay, this one's got some pretty crazy potential. These changes add a very easy and comfortable method of adding audio when equipping and unequipping clothing items. The whole system is implemented and pretty easy to use from here on, and will work with every single subclass of "clothing".

image
*Pictured the variables in the "clothing" class, which will then work in any of its subclasses.*

These changes also add some sounds already! I made some velcro and vest swooshing sounds to the majority of the vests that look like they would use some sort of velcro straps.

image
*Pictured are the vests that now have audio. Also pictured is a security smooth boy.*

Finally, I also did a very small rebalance of how long it takes to equip and unequip armor. No more taking off a full riot suit in 1 second. It's been rebalanced so that every single armor vest takes 2 seconds at least to take off and put on, and all "full" armor vests take 3 seconds to take off and put on, since they have additional parts like elbow and knee pads.

**Important things to note:**
- In order for the audio to play, the variable `equip_delay` **MUST BE ABOVE 0**. It is set to 0 by default and must be set to any number above it, or else the audio code will not get a chance to run.
- Audio files must be directed using `'[patch to .ogg file]'`, not `"[path to .ogg file]"`.

image
*Pictured is an example of the audio code used on one of the vests.*

Github doesn't let me upload audio files here, so uh... I'll just post them in the discord I guess.

## Why It's Good For The Game

It makes it easier for future implementations of audio when equipping or unequipping clothing, and it just makes the whole game more immersive. Additionally, the balance changes make it so the audio files play correctly as well as make it a little more tough to put on or take of armour. It IS a massive vest after all, that takes time to put on and strap up correctly.

## Changelog
```changelog
balance: Balanced how long it takes to put on and take off vests. All vests now take 2 seconds, and full vests take 3 seconds.
soundadd: Added audio files for putting on and taking off armor vests.
code: Added an audio system for clothing.
```


Jamini (approved):
If it's too hash I suppose we can revert it. Approved so we can get it in highpop.

PR #599, opened 2021-07-12
Wielded weapons no longer embed

NotRanged:
## About The Pull Request

The vote passed pretty resoundingly so here it is. It's also done in a more optimized way than Eris did it.

## Why It's Good For The Game

I even tested this :)

## Changelog
```changelog
balance: Wielded weapons can no longer embed.
```


Michiyamenotehifunana (approved):
Very elegant code, 10/10 would recommend.

PR #601, opened 2021-07-15
Roundstart Slimepeople

Michiyamenotehifunana:
## About The Pull Request

qww1bVLgbM

This PR adds roundstart slimepeople. The only 3 abilities they inherit from xenobio slimepeople are: Gibbing on death (core implants are retained), noslip, and limb regeneration. They otherwise lack the super stungloves and need to actually breathe. Their stats have also been normalized to be in-line with other roundstart species. They are not friendly to actual slimes.

pv08HHWMMk

One issue with the sprites is that they come precolored, meaning you can't truly get the color you want. This will be fixed in a later PR if this goes through. 

## Why It's Good For The Game
More possibilities for characters, more roleplaying potential.

## Changelog
```changelog Toriate
add: Roundstart slimepeople now available in the character creator!
```


IrkallaEpsilon:
Suggest to rename them to Prometheans or something, lesser sounds very judging tbf even though they are a downgrade from xenobio slimepeople

fludd12:
If this does indeed go through I plan on doing a complete mechanical rework, as a bit of a heads up.

ShadzFen:
I'd like to see more details on how they work mechanically, cus the conversation we talked on the discord regarding it was a mess. Myself and many more are pretty unfamiliar with original slimepeople to begin with, since xenobio is basically an untouched job. 

- What are their stats in comparison to humans? 
- Is limb regeneration and noslip their only advantageous skills?
- How would they work regarding Erismed? IE, exactly what organs do they have and what medical methods work / don't work, etc etc. (e.g., is it like TG slimes that die in cryotubes? do they have maybe 120 health like unathi compared to humans 100? etc)

Michiyamenotehifunana:
They take 10% more brute damage in exchange for taking 10% less burn damage, where the xenobio ones have far larger numbers over 30%. They take 3x more damage from electric shocks. They also lack the instant taser melee attack, and instead just do regular punches and kicks. They have more health in order to compensate for the fact that they instantly gib on death.

Limb regeneration and not slipping should indeed be their only true advantages.

Slimes lack conventional internal organs such as kidneys and livers, but still have muscles and nerves which function as normal. Their only true organ is the slimecore, located in their head. Removing the slimecore or otherwise decapitating a slime will instantly gib them. There does not appear to be any special interactions regarding antitox stuff (and by extension, cryotubes) and slimes, but I am not confident enough to confirm it either way.

Jamini:
Okay, time to review this.

Through the lens of coding, looks fine. Everything is in place.

Through balance, I see a few concerns:

Neutral
- 1.1 Brute mod for 0.9 Burn mod. Fair and balanced

Pros:
- NO_SLIP - I know what you are getting at with this Tori, but this is a very large advantage. This makes you completely immune to water, soap, clown pdas, and space lube. 
- regenerate_organs() - Very, very powerful. Even with the very low health on slimeperson limbs. For most people losing a limb is an extremely big deal, and this completely negates that.
- Lack of internal organs - Very, very powerful as well. Fewer things to get damaged means much less reliance on medical. This is going to make them *incredibly* durable. way more-so than even the erismed2 mobs. Infections are much less lethal to them, and they will suffer less slowdown due to hunger as internal organ repair is the #1 cause of nutrition drain.

Cons:
- Gibs on death - Very dangerous in maint. AED is ineffective to revive them.
- x3 shock damage (this is actually 2.7 because they have a 0.9 burn resist)
- Lower limb health. Meaning it is more likely that they will need to regenerate.

Right now the death gib makes up for the regeneration, however there are no downsides for noslip and lacking internal organs to damage.

Flat out, this species needs more downsides to compensate for their upsides. Even after the nerfs you've applied. I would suggest the following:

- A slowdown of 1, equal to permanently wearing Galoshes. This can be compensated for by using hyperzine or having hydraulics installed.
- Self-cannibalism. Should a slime-person run out of nutrition, they will get a short warning before one of their external limbs cannibalizes itself. restoring nutrition to the slime-person.
- A heart and lung equivalent.

Michiyamenotehifunana:
Slimes no longer get noslip, they get 25% more blood, and they get a heart and eyes.

In addition, they no longer use a snowflake species icon, instead using the base human sprite but with transparency code fudged in.

Jamini (approved):
Tentatively approved. VERY tentatively approved.

PR #610, opened 2021-07-22
This can be safely ignored as it changes nothing

Jamini:
fixing some wonky CLF stuff


Michiyamenotehifunana (approved):
Why isn't our repo set up to automatically deal with CRLF bullshit.

ShadzFen (commented):
I have no idea what I'm writing but I trust the fact that you tested it and it works.

PR #619, opened 2021-07-30
Elevator bug fixes!

ShadzFen:
## About The Pull Request

This PR fixes one of the main issues the elevator had: Not deleting IDs.

- The elevator will now check to see if your ID is the captain's spare, and if it is not, it will delete the ID correctly without dropping it like it used to.
- Any reserved items the player may have on them will now get put into a Lost and Found unit on the outside of the elevator rather than dropped to the floor.
- The elevator will now clean up all items dropped inside it (except for the intercom obviously) and put them into the Lost and Found unit outside the elevator. Do keep in mind that the clean up process is called when the elevator is called DOWN, i.e., when someone presses the call elevator button outside the elevator.
- Additionally did some commenting and made some of the code a little cleaner and easier to read.

This is part 2 of the elevator addition. Part 3 will come soon, and will include some small easter eggs, an emag option for the elevator and POSSIBLY (if I can figure out how to code it without pulling my hear out) a system that allows players to join via coming down the elevator in order to improve immersion. More info on this will probably be discussed during a future emojicracy, as this will possibly make spawn in times via the elevator slower, or potentially not possible if the elevator is broken IG.

**| | | NERD STUFF FOR JAMS TO READ | | |**

You need to map in a wall and a console to the north of the elevator, and give the console a **specific Y-nudge of 32** in order to follow the same standard as the other panels on the ship.

image

image

Here's an example of what it should look like!!!!!

image

## Why It's Good For The Game

People won't be able to nab other people's dropped IDs and use them. The code is easy to read and understand.

## Changelog
```changelog
add: A cleanup system for the elevator that makes it not be a disgusting mess several hours into a shift.
fix: Fixed a bug where ID cards were getting dropped rather than being deleted.
spellcheck: Capitalized the Elevator area, so that it is capitalized similarly to all other areas.
code: Commented a lot of the elevator code for easy reading and understanding.
```

### DIFF
--- a/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
+++ b/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
@@ -204,14 +204,13 @@
 	var/obj/machinery/computer/cryopod/elevator_storage/lostFound
 
 /obj/machinery/button/remote/elevator_call_button/trigger()
-	for(var/obj/machinery/computer/cryopod/elevator_storage/storage in range(usr, 5))
-		lostFound = storage
 	for(var/obj/machinery/door/airlock/D in range(src, 5))
 		if (D.id_tag == id && D.locked) //Gotta check if the airlock is locked. This is how we know whether the elevator is at our floor or at the dorms.
 			for (var/obj/item/W in get_area(D)) // For every item inside the elevator area... (The area that the elevator airlock is in)
 				if (!istype(W, /obj/item/device/radio/intercom)) //Stop eating my fucking intercom you cockwaffle
-					W.forceMove(lostFound) // Move the items to the lost and found in order to clean up the elevator.
-					lostFound.frozen_items += W // And add them to the inventory list.
+					for (lostFound in range(usr, 5)) // Get any Lost and Found units within 5 tiles...
+						W.forceMove(lostFound) // Move the items to the lost and found unit in order to clean up the elevator.
+						lostFound.frozen_items += W // And add them to the inventory list.
 			for(var/obj/machinery/button/remote/elevator_panel/B in range(src, 5))
 				if (B.elevator_moving == FALSE) //If the elevator is no longer moving I.E., the timer it sets has finished, then we are ready to call it back down.
 					if(world.time - cooldown_timer > cooldown)
--- a/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
+++ b/zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
@@ -4,26 +4,43 @@
 
 /obj/machinery/button/remote/elevator_panel/proc/despawn_passenger(var/mob/living/user)
 	var/mob/living/passenger = user
-	//var/mob/living/carbon/human/passengerplayer = user
+	var/obj/machinery/computer/cryopod/elevator_storage/lostFound
 
-	for(var/obj/item/W in passenger)
-		if (istype(W, /obj/item/modular_computer/pda))
-			var/obj/item/modular_computer/pda/found_pda = W
-			found_pda.eject_id()
-		else if (istype(W, /obj/item/weapon/implant))
-			var/obj/item/weapon/implant/found_implant = W
-			found_implant.uninstall()
-			qdel(found_implant)
-		else if (istype(W, /obj/item/weapon/storage || /obj/item/clothing/suit/storage))
-			for(var/T in preserve_items)
-				for (var/obj/item/w_storage in W.contents)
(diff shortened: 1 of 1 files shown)


zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -4,26 +4,43 @@
...
-			found_implant.uninstall()
-			qdel(found_implant)
-		else if (istype(W, /obj/item/weapon/storage || /obj/item/clothing/suit/storage))
-			for(var/T in preserve_items)
-				for (var/obj/item/w_storage in W.contents)
-					if (istype(w_storage, T))
-						passenger.drop_from_inventory(w_storage)
+	var/obj/machinery/computer/cryopod/elevator_storage/lostFound
+
+	for(var/obj/machinery/computer/cryopod/elevator_storage/D in range(usr, 5))
```
Michiyamenotehifunana, line 9: And what if you want to map the console somewhere further than 5 tiles away?

ShadzFen: It won't be 5 tiles away, it will be outside the elevator, as I showed in the pictures.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -158,10 +201,17 @@
 	desc = "An button used to call the elevator down to the current floor."
 	var/cooldown = 24 SECONDS
 	var/cooldown_timer
+	var/obj/machinery/computer/cryopod/elevator_storage/lostFound
 
 /obj/machinery/button/remote/elevator_call_button/trigger()
+	for(var/obj/machinery/computer/cryopod/elevator_storage/storage in range(usr, 5))
```
Michiyamenotehifunana, line 207: Ditto, what if someone wants to map this further than 5 tiles from the thingy?

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
```diff
@@ -158,10 +201,17 @@
...
+		lostFound = storage
```
Michiyamenotehifunana, line 208: This should be in initialize() or something, this makes it constantly update this var every time someone hits the button, which misses the whole point of making it a var to begin with.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
Jamini: Tori, why would the button for an elevator be more than five tiles away?

Michiyamenotehifunana: What if someone makes a crazy elevator design that necessitates it!

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
Michiyamenotehifunana: It'd still allow for a lot more mapping flexibility if it iterated through the machines subsystem's machines list to find the console instead of having everything be within 5 tiles.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: It won't be more than 5 tiles away, it's right next to the elevator...... if they want to map it to be further away, they can change the damn range themselves lmao

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: Really not an issue... Again, if they want to map it that absurdly, they can do so and change the range

Michiyamenotehifunana:
RANGE ARGUMENTS AND MAPPING CONSTRAINTS ASIDE, both the button and the panel has a var for storing the console's ref, but it isn't used at all since everytime the thingies are used, it searches for the console again.

Ideally, it should only search for it on initialize, then just use that var without trying to relocate it every time it's used again.

Jamini:
Shadz, I know how to nudge items on walls. Seriously, BRUH.

zzzz_modular_occulus/code/game/machinery/dorms_elevator.dm
ShadzFen: fixed

Jamini (approved):
Should be "good enough" for now. Shadz already mentioned he is doing a third pass later to add more functionality.

PR #622, opened 2021-08-03
Mob Agony Coefficients, new mobs, and the EES SED

Michiyamenotehifunana:
## About The Pull Request

UPDATE: The Synthetic Entity Detector has been added, it's a device that spawns robots in random areas in maint and has a cooldown of 20 minutes. It is an EES thing, only reporting its results on engineering comms.

StrongDMM_xFYq6v6Ydc

First and foremost, this PR adds a new modifier to roaches: The Agony Coefficient. It affects how much damage mobs take from rubbers and other pain damage sources. Mobs default to a value of 1, meaning they take full one-to-one damage.

Mobs like panzer roaches have been given values as low as 0.8, meaning they take 20% less damage from pain sources. Others like the Seuche and Gestrahlte have 1.2 instead, owning to their (probable) squishiness. The Stranger has also been given a coefficient of 0.5.

In general, all robotic mobs also have lower coefficients. Roombas have 0.8, while combat drones have 0.5.

This PR also adds new mobs.

RoombattlerPreview2
The combat roomba comes in two variants, one double-shielded one that has an agony coefficient of zero, and one that shoots weak rubber bullets in bursts. They only spawn when the EES SED is used, and will drop appropriate loot.
dreamseeker_AkEMsSDkA3

dreamseeker_SoVWnKxlJA

dreammaker_fBkdCWOsOn
SPUD units are generally friendly mobs that attack roaches. They will melt if a swarm gets on them, however. There is also a blood-red infiltrator variant that is hostile instead.

Sprites made by me, exclusively licensed for use only on whatever servers I decide, such as this.

## Why It's Good For The Game
More immersion from robots not super easily dying from rubber bullets, more variety in combat tactics, and also more variety in enemies.

## Changelog
```changelog Toriate
add: A new shipment of robots from Bosstown Dynamics Incorporated has arrived on board! Unfortunately, their programming was interrupted by an untimely ion storm, causing them to run amok. Find them in maintenance today!
balance: Mobs now have agony coefficients that modify the amount of pain damage they take, making rubber bullets more or less effective against certain enemies. Robots in general take less pain damage.
```


Jamini:
I... do not see why the roombas would be roach-aligned? That goes against other mob faction concepts, and doesn't really fit with the lore to boot.

I'm also really not sold on reducing agony to brute conversions. The few mobs we do that take reduced/no damage from rubbers (slimes) are absolutely ass to fight. This feels like it would be extremely unfun, especially as Panzers already have extremely high health values.

Rather than spawning spuds, wouldn't it make more sense for us to build them (something for robotics to build cheaply at the start of the round?) to combat roaches?

ShadzFen:
Right so if I understood properly what was discussed on the Coding Discussion channel:
- Roombas are gonna be high accuracy, high damage mobs that spawn in clusters or 3-4-5.
- They will replace the use of the Reality Complicator by instead spawning a group of them in a random, announced location, or perhaps only announced over the engineering radio? Just to keep it more... private, since it *is* for the sake of giving engineering more stuff to do.
- I think we agreed on having them spawn in high loot locations in maintenance too, but I dunno if we should do that other than the bridge maintenance personally.
- They will no longer be roach aligned for the sake of making them seem more hostile. (The example used was walking into a room to find 5 combat roombas driving around among roach corpses and pools of roach guts being much more ominous than an empty room)

If I missed anything, feel free to add it in an extra comment.

Michiyamenotehifunana:
Roombas are no longer friendly to roaches, EES SED has been added to facilitate their spawns without interfering with anything, special infiltrator variant of the syndie SPUD has been added for the only roach aligned thing.

Main issues should be resolved.

maps/CEVEris/_CEV_Eris.dmm
```diff
@@ -105942,6 +105944,11 @@
 /obj/structure/multiz/ladder,
 /turf/simulated/floor/plating/under,
 /area/eris/maintenance/section1deck4central)
+"meF" = (
```
Jamini, line 105947: What is this?

zzzz_modular_occulus/code/modules/mob/living/simple_animal/hostile/nanoblob.dm
```diff
@@ -37,6 +37,8 @@
 
 	environment_smash = 1
 
+	agony_coefficient = 0.8 // rubbers aren't going to hurt a blob of goo very much
```
Jamini, line 40: Wouldn't this be the other way? If anything the way that LTL works (larger contact area at a slower speed) actually hurt a geletanous creature more?

Michiyamenotehifunana:
Changes done.

ShadzFen (dismissed):
Everything looks okay, mostly modular, all the new changes and additions seem to be in order. Looking forward to actually seeing this one in the game!

Jamini (dismissed):

Jamini (approved):

PR #623, opened 2021-08-03
red shemaugh

Michiyamenotehifunana:
## About The Pull Request

Aseprite_irU2kuNLia

This PR adds the red shemaugh to the loadout.

## Why It's Good For The Game

fashion

## Changelog
```changelog Toriate
add: Added red shemaugh
```


Jamini:
Wouldn't it be better to just make the Shememah colorable?

ShadzFen:
Yeah I was thinking that too but forgot to comment on it. Might be a better idea to just make it colourable, or add a larger selection of colours than just 1 more colour.

ShadzFen:
So what's it gonna be? Are you gonna make it colourable or just add a red shemaugh for now? We'll probably end up getting people saying "Why are there no other colours" if we don't add more colours, and I do think making it colourable would be way cooler. Not to mention the code for making it colourable could then be copy pasted onto other stuff, both existent and future items in the loadout. Just something to think about. Personally I'd love to see colourable, even if it's just a box where you input a hex colour.

Michiyamenotehifunana:
Unfortunately out of scope for this PR, we can easily add colorable clothing some other time considering it's basically been done to death by this point.

Jamini:
The code for colorable items already exists. It's a simple manner of implementing it. We have a large number of items that can be recolored already.

ShadzFen (approved):
Oki doke. We should definitely consider doing it at some point though. I'd love to see colourable loadout items similar to how we have custom loadout items.

PR #630, opened 2021-08-05
Fixes event refund code

Jamini:
## About The Pull Request

fixed a single conditional that  was causing multiple partially-failed roleset events from triggering one after another.
## Changelog
```changelog
fix: pro-rata point refunds should now work.
```


ShadzFen (approved):
Oh cool. Lets get it merged then

PR #631, opened 2021-08-07
Reverts cat ears

cebutris:
## About The Pull Request

As the servers authority on catgirls (I'm pretty sure only one other person plays those), I can say with near certainty that the new cat ears suck. This brings them back to their original state

## Why It's Good For The Game

Better looking catgirls are better catgirls


Jamini:
Tajara use these ears too. Get me the original file, don't revert it.

PR #639, opened 2021-08-12
Fixes White Nail Sprite Bug

razekpraxis:
## About The Pull Request

White Nail would briefly have no sprite when fired due to ammo not being a multiple of 0.25. Adds a rounding check to fix this.

## Why It's Good For The Game

Fixes a rather glaring bug that I should have caught when I was doing final tests. 

## Changelog
```changelog
fix: fixed sprite bug with White Nail
```


ShadzFen (approved):
All good.

PR #644, opened 2021-08-13
Update blattedin.dm

Jamini:
## About The Pull Request

Prevents Blattedin from getting stuck in internal mob reagent processes

## Why It's Good For The Game

REM is not supposed to be multiplied by fractional numbers. Going too low in your metabolism rate breaks certain chems. This is a hotfix for blattedin specifically.

Generally you can go to REM * 0.25 before things get really wonky, but even then we really shouldn't be reducing metabolism AT ALL. And whoever changed this initially was extremely foolish

## Changelog
```changelog
fix: Blattedin no longer should get stuck processing in mobs.
```


Michiyamenotehifunana (approved):
Looks good.

PR #646, opened 2021-08-15
Makes the last shelter work

Michiyamenotehifunana:
## About The Pull Request
Now cruciform bearers will get gibbed and have their cruciforms teleported to the Last Shelter instead of just being stuck in space.

dreamseeker_Vbaui9CCKJ
dreamseeker_vGbyjJx3rd

It notifies all cruciform bearers when that happens.

## Why It's Good For The Game
This is what everyone thought the Last Shelter would have done, rather than being a glorified ghost role spawner.

## Changelog
```changelog
add: Last Shelter now recovers the cruciforms of people lost in space.
```

PR #660, opened 2021-08-18
Revert "Makes the last shelter work"

Jamini:
Reverts Occulus-Server/Occulus-Eris#646


Michiyamenotehifunana (approved):
emergency approval due to game breaking bug

PR #670, opened 2021-08-22
Rounds off Union Exports

Shadow-Quill:
## About The Pull Request

Deletes pennies from Union exports. This uses round() which in byond, is just flooring something. (Read: Rounding down.)
The rounding only happens right before the credits are added to the Union account, so there won't be any losses if three items have XXX.4 credit value.

## Why It's Good For The Game

Pennies shouldn't exist.

## Changelog
```changelog
tweak: Union exports will no longer put pennies (less than whole credits) into the account.
```


Jamini:
NOOOO MY PENNIES - approved

Jamini (approved):

PR #671, opened 2021-08-23
Fixes some tool damage issues and other things

Jamini:
## About The Pull Request

This PR fixes tools losing damage mods when being wielded or turned on or off. It also touches a bunch of other smaller things. Please see the changelog

## Why It's Good For The Game

Damage mods will be more consistent now.

## Changelog
```changelog
tweak: fertilizer bottles now are eris bottles and should be able to have variable transfer amounts
fix: tools with force_wielded should no longer lose damage mods when being wielded
fix: tools with force_switched_on should no longer lose damage mods when being turned on.
fix: cleaning the bioreactor glass will no longer make it impossible to see through.
del: removed an invisible hat in the HoP locker.
icon: lootbags should now properly display.
```


Michiyamenotehifunana (approved):
Looks good, sanity checks check out, nothing egregious in the rest of the stuff.

PR #677, opened 2021-08-24
Elevator 3.0

ShadzFen:
## About The Pull Request

Elevator 3.0 was supposed to be my last elevator PR but given that I'm bad at stuff, the next part of it will have to come in its own PR. This one is relatively small, but it makes some QoL changes.

The elevator panel itself now handles whether you go up or down, rather than having 2 different buttons for it, which was kinda dumb. It also now plays the correct audio when going down, rather than playing the audio you would hear outside the elevator while inside.

If a player spawns in, lets say by accident, and the elevator is on deck 6 (IE, the airlock is closed and locked), selecting "Deck 6" and leaving the shift will instantly despawn, rather than go up again, which wouldn't make much sense. 

When the elevator reaches a floor, it now announces what floor it has reached. Very small addition but I think it's nice uwu

This is the new message that shows when you click the elevator panel.

image

The old message still shows when you select Deck 6.

image

The code has been cleaned up to look much nicer now too, I think at least. Looks easier to read. 

Finally, removed the old redundant button and changed the spawn location 1 tile north so that it doesn't put the elevator's ship intercom into someone's pocket due to spawning on top of the elevator coms. Dumb bug.

image

## Changelog
```changelog
add: New system for going up and down in the elevator, as well as a basic floor announcement system.
fix: Incorrect audio playing when going down in the elevator.
fix: Spawning in the elevator will not longer randomly put a ship intercom in your pocket.
code: Cleaned up going up and going down to make it easier to read and modify in the future, if need be.
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -106709,12 +106709,7 @@
 /turf/simulated/floor/tiled/white/gray_platform,
 /area/eris/medical/surgery)
 "nIM" = (
-/obj/machinery/button/remote/elevator_call_button{
-	dir = 4;
-	id = "ElevatorDorms";
-	name = "Elevator Exit Button";
-	pixel_x = -26
-	},
+/obj/landmark/join/late/dormitory,
 /turf/simulated/floor/carpet/turcarpet,
 /area/eris/elevator)
 "nPo" = (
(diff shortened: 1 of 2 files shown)


Jamini (approved):
Nothing egregious. I retain the right to mercilessly use Shadz' cutout if this breaks in a hilarious way.

PR #685, opened 2021-08-29
unfuckles carpet crates

Zenitheevee:
## About The Pull Request

Makes the carpets in carpet crates actually work.

## Why It's Good For The Game

fixes a fucky wucky i made OwO

## Changelog
```changelog
fix: fixes carpet crates by tying the amount to the crate, and not making a child that i thought would inherit the parent tile type, but for some reason didn't, and so i had to make this innane PR fixing an issue i should have tested, but had no idea it wouldn't have worked except for the few times i DID accidentally come acrossed it but never checked because i'm an idiot.
```

### DIFF
--- a/code/game/objects/items/stacks/tiles/tile_types.dm
+++ b/code/game/objects/items/stacks/tiles/tile_types.dm
@@ -95,31 +95,7 @@
 /obj/item/stack/tile/carpet/oracarpet
 	name = "orange carpet"
 	icon_state = "tile_oracarpet"
-//occulus additions start for carpet crate packs
-/obj/item/stack/tile/carpet/cpack
-	amount = 30
 
-/obj/item/stack/tile/carpet/bcarpet/cpack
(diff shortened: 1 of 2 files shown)


Jamini:
Is this tested and working Zeni? Can we get a screenshot of this functioning on a test server please?

Zenitheevee:
So, Because images wont be the most helpful in this case, i can.. kinda explain what the issue was.
Basically, when i made the crates, i wasn't the most aware of the "Amount" variable, how it interacted, such and such, so i, _in my infinite wisdom_, made a child of the carpets with amounts tied to them instead. 
This, in my mind, Would resolve easily because it would call back to the parent's tile variable... _But Byond being Byond_ didn't think so.
image
Anyways, Image of the carpet and textbox to prove no game panel was used.

Michiyamenotehifunana (approved):
Looks good!

PR #690, opened 2021-09-02
Logging for the storyteller

Jamini:
## About The Pull Request

Adds a log and notification when someone messes with point pools in a storyteller

## Why It's Good For The Game

Tracking this stuff is good.

## Changelog
```changelog
admin: point pool modifications by the storyteller panel are now logged
```


ShadzFen (approved):
Ayy you actually used that one github tag that never gets used correctly, nice. This is good. Unmodular addition, labeled in code. Good to go

PR #691, opened 2021-09-03
Medical QoL improvements.

ShadzFen:
## About The Pull Request
Just a few small adjustments to improve the use of certain medical items.

- Gauze and Ointment now have 10 uses instead of 5.
- Advanced brute kits and burn kits now have the same chance to not waste materials during surgery as they do when healing injuries.

## Why It's Good For The Game
This change will give players that have low BIO stats the possibility to share a standard First-Aid kit, rather than using one up entirely on themselves. It will also help doctors with high BIO to save materials on the valuable advanced kits during extensive surgery.

## Changelog
```changelog
balance: Roll of Gauze and Ointment increased from 5 to 10 uses.
fix: Advanced kits will now properly apply the chance to not waste materials during surgery mending.
```


Jamini (approved):
Might make it too hard to run out f medical supplies, but w/e we have loads of them anyway. Approved.

PR #696, opened 2021-09-05
Porting cigarette things from upstream that we missed

Michiyamenotehifunana:
## About The Pull Request

In this PR: janky Vape masks (no actual vapes), more ciggies, and giant ciggie boxes. And also a matchbox sprite regression but upstream didn't care so we can deal with that later.

dreamseeker_uii7mRTFAH
dreamseeker_yKif26nwO8

Vape masks need to be opened with a screwdriver before they can be filled or have their batteries swapped.

For some reason the new stuff didn't show up in the vending machines when I tested it, but I'm assuming I messed up something when I compiled it. Except now I realized we may have a modular override somewhere just as I type this out.
## Why It's Good For The Game

More content is good.

## Changelog
```changelog Toriate
add: More cigarette types, jumbo ciggie packs, and vapes from upstream! 
```


Jamini (approved):
I recall skipping this Pr intentionally for some reason, but I can't think of why. Merging and hoping nothing breaks.

PR #703, opened 2021-09-10
Long Range Scanner and Trade Station fixes

Jamini:
## About The Pull Request

This PR makes it so that our long-range scanners have a variable range. At the cost of exponentially more power draw, you can increase the range up to 8 tiles (the maximum viewrange)

Additionally this cleans up and tweaks the offers provided by a number of trade stations. Check your local trade program console for info!

## Why It's Good For The Game

This should make exploring easier for bridge staff, as well as make fulfilling offers more lucrative for The FTU

## Changelog
```changelog
balance: Trade Requests from outside stations now cap at 10,000cr instead of 20,000
balance: Mines, EMP mines, and Bear Traps now have value
balance: The Suit-Up station no longer starts discovered
balanced: made markups for a number of stations higher.
tweak: Long Range Sensors can now scan variable ranges, drawing exponentially more power to do so.
tweak: Touched on every trade station, cleaning up descriptions and adding a number of potential offers
fix: removed stack-based offers from trade stations. As the current code doesn't account for stacks
fix: removed the trade beacons from centcomm
```


ShadzFen (approved):
Looks cool. The names and introductions could have a little more variety to each, so we don't have one trader introducing themselves the same way over 10 shifts in a row, saying that they're about to leave or whatever. But that can be added at a later date or something. 

Good to go

PR #707, opened 2021-09-14
Medical supply balances

RadiantFlash7:
## About The Pull Request

Balance talked about in discord coding channels:
Drops standard kit amount for oxy and fire kits to 2.
Removes Hypokits from table, leaves them within the vendor.
Removes individual hypos and cartridges from vendors, Moves them to chemistry storage.
Also replaces CMO locker with CMOcculus locker
## Why It's Good For The Game

balance. Wasn't my idea, but it's a suitable compromise.
## Changelog
```changelog
del: Removes 2x fire&oxy medkits from map
del: Removes hypokits from table in medical storage,
tweak: moves Apollo and Apollo cartridges from vendor to chemistry
fix: Replaces CMO locker with the correct CMOcculus locker.
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -73284,10 +73284,7 @@
 /turf/simulated/floor/tiled/steel/gray_platform,
 /area/eris/medical/chemstor)
 "dsK" = (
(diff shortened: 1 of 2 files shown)


ShadzFen (approved):
Simple map changes. All good

PR #730, opened 2021-10-04
Re-adds cargo trains, compacting piles by hand

Jamini:
## About The Pull Request

This PR re-adds the cargo tug and trollies to the map.
At current there are no means to order more, and they still need a new sprite
It also rebalances a few common tools and adds a way to make scrap cubes that isn't an exosuit

## Why It's Good For The Game

More trolly fun for cargo, more scrap cubes. Makes shovels a bit less redonk in qualities and buffs true hammers to be in line with other tools

## Changelog
```changelog
balance: Regular hammers to 30 hammering from 20
balance: Homewreaker hammer to 30 hammering from 15
balance: Power hammer to 50 hammering from 30
balance: Hammering quality removed from all shovels
fix: buckled mobs can no longer use ladders
balance: Trollies are not limited in what can be loaded.
map: adds a trolly engine, key, and three trollies to cargo on the middle level
```

### DIFF
--- a/code/game/objects/items/devices/lighting/toggleable/flashlight.dm
+++ b/code/game/objects/items/devices/lighting/toggleable/flashlight.dm
@@ -255,7 +255,7 @@
 
 			user.visible_message(SPAN_NOTICE("\The [user] directs [src] to [M]'s eyes."), \
 							 	 SPAN_NOTICE("You direct [src] to [M]'s eyes."))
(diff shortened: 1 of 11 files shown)


ShadzFen:
Did you fix the layering issue with the tugs? Specifically when someone rides it.

ShadzFen (approved):
Look fine

PR #748, opened 2021-11-01
A few small tweaks here and there.

Jamini:
## About The Pull Request

Fixes a few small bugs

## Why It's Good For The Game

Buxfixes and small additions.  Stuff everyone likes.

## Changelog
```changelog
add: Added "You will be Okay" to the jukebox under cc_by_sa_4_0 license
fix: The Crypt Access Updater no longer will delete access off ID cards.
fix: The Gemini and Glasgow disks are now visible, and should print their respective guns
fix: The skill implanter should once again be visible.
```


code/modules/soulcrypt/access_update_tool.dm
```diff
@@ -55,7 +55,8 @@
 			return
 		//Else, we overwrite the soulcrypt's current access with the access on the ID card.
 		crypt.access.Cut()
-		crypt.access = card.access
+		for(var/x in card.access)
+			crypt.access += x
```
fludd12, line 59: An alternative would be to do [crypt.Copy()](http://www.byond.com/docs/guide/chap10.html), though that may require length arguments. Either way, this works to de-reference the list, assuming the vars in access are not also references.

fludd12 (approved):
Looks good to me, left one note about a shorter method of de-referencing lists.

code/modules/soulcrypt/access_update_tool.dm
Jamini: Copy() did not work when attempted. It requires some conversion to Lazylen first from what I can tell.

PR #761, opened 2021-11-19
Adds Two New Nanite Weapons (Black Talon, Grey Claw). Code adjust to White Nail.

razekpraxis:
## About The Pull Request
**THIS WEEK AT NANOTRASEN (TWANT)**

This week at Nanotrasen we've expanded our Nanite family!

Our engineers have been hard at work to bring you not one, but two brand new weapons to the Nanite line. This update also brings with it some much needed housekeeping in for messy code. Each new weapon comes with a suite of sounds, with White Nail receiving a small audio adjustment.

### Black Talon

Nanite PR In-Game Talon

The Black Talon is a condensed version of the White Nail. It'll cost you 1900 upfront, and needs 20 uncapped nanites to build. It can also fit in a pocket, or a holster! It uses the same ammunition as the White Nail, though suffers an effective 0.2 negative damage modifier meaning its overall damage output is lower.

- Fires in 2 round bursts
- Holds 4 bursts maximum
- Slightly quicker recharge time vs. White Nail

https://user-images.githubusercontent.com/17753168/142628967-46b17fe8-9534-404a-b8d1-f6fdf46421d8.mp4

### Grey Claw

Nanite PR In-Game Claw

The Grey Claw is a heavy duty shotgun variant of the White Nail.

- Fires single shot
- Can switch between slug and pellet mode at will without losing current ammunition count
- Holds a maximum of 2 shots
- Slow recharge rate vs White Nail

https://user-images.githubusercontent.com/17753168/142628988-67feea07-9b8c-4b99-9a07-ada5ad57b109.mp4

### White Nail

White Nail also received some small adjustments alongside a general code improvement.

- **White Nail's Credit Cost is down from 6000 base to 3500.**
The gun's very expensive to buy. Prohibitively so. Cost has been reduced while still keeping it on the upper tiers of expense for a weapon. 

- **Recoil build up reduced to 5 (was 7).**
General feedback on weapon was that the recoil build up for a rifle like this was too high. 

- **Number of Uncapped Nanites required to build reduced to 24 (was 25)**
Functionally this change does nothing, but now the number is divisible by the amount of nanites you get from a single Kraftwerk roach. (You need 6 roaches!)

- **Adjusted White Nail flavour text slightly.**

- **Changed White Nail wield sound**
I wasn't in love with the way it sounded. Now I am!

### General Changes
More generally the generate_ammo() proc for all three weapons has been improved from the original release. 

- The code has been adjusted to work with any sized magazine, rather than being hardcoded for the White Nail's magsize. 

- A similar change has been applied to the code to change the sprite based on how full the magazine is. 

- Also the message that plays when the gun is full will now use [src.name] in the string rather than, again, being hardcoded for White Nail. 

All new weapons and magazines have been added to the R&D tree. All three weapons are available under the same tree node, meaning only one unlock is needed. Cost for building the weapons remains a limiting factor on the availability of the guns.

A massive thank you to @Michiyamenotehifunana for the sprite work!!

## Why It's Good For The Game

Fixing up the previous code to be less hard-coded is a general good practice and is something I've used to improve my skills as a programmer personally (though I know this isn't game-relevant)

Adding two new guns to the line-up will introduce more diversity and fill more gaps. Feedback was that a single three-burst rifle was interesting but not enough to get some people to use it. These two new guns will hopefully fill niches of their own. Black Talon being a lightweight but underpowered option, Grey Claw being heavy duty and methodical.

## Changelog
```changelog
add: Added Grey Claw shotgun
add: Added Black Talon handgun
balance: Adjusted recoil build up for White Nail
balance: Adjusted cost for White Nail
tweak: adjusted code to be less hardcoded for a single gun and flexible for any magsize
soundadd: added Grey Claw fire and wield sounds
soundadd: added Black Talon fire and wield sounds
soundadd: added new White Nail wield sound
sounddel: removed old White Nail wield sound
imageadd: added spriteset for Grey Claw
imageadd: added spriteset for Black Talon
spellcheck: Changed White Nail flavour text slightly
```


zzzz_modular_occulus/code/modules/projectiles/guns/projectile/rifle/whitenail.dm
```diff
@@ -33,7 +34,7 @@
 	START_PROCESSING(SSobj, src)
 	update_icon()
 
-/obj/item/weapon/gun/projectile/automatic/whitenail/Process() // increments counter by one per tick until it reaches 
+/obj/item/weapon/gun/projectile/automatic/whitenail/Process() // increments counter by one per tick until it reaches gun's rebuild_time var
```
fludd12, line 37: I know it's rather late to be commenting this since the White Nail is already in the code, but we may want to take another look at how it handles ammo generation and processing.

A slightly more effective method in this case would be to only have the gun process when it is not full on ammo - stopping when it gets full, and starting again after being fired.

zzzz_modular_occulus/code/modules/projectiles/guns/projectile/rifle/whitenail.dm
```diff
@@ -42,16 +43,17 @@
 		generate_ammo()
 
 /obj/item/weapon/gun/projectile/automatic/whitenail/proc/generate_ammo() // generates ammo in the loaded mag
```
fludd12, line 45: Similarly, this proc can be reworked, generating up to the max ammo count and no higher using `min(ammo_magazine.max_ammo - src.get_ammo(), 3)`.

zzzz_modular_occulus/code/modules/projectiles/guns/projectile/rifle/whitenail.dm
```diff
@@ -42,16 +43,17 @@
...
 	if(src.get_ammo() >=  ammo_magazine.max_ammo) 
 		return FALSE
 	
-	for(var/i in 1 to 3)
+	for(var/i in 1 to rebuilt_bullets)
 		ammo_magazine.stored_ammo += new ammo_magazine.ammo_type()
 
 	update_icon()
 
 	if(src.get_ammo() >=  ammo_magazine.max_ammo)
```
fludd12, line 55: Here is about where I'd put STOP_PROCESSING.

fludd12 (approved):
Everything looks pretty good! Certainly in line with the already very nice White Nail code.

I'm approving rather than requesting changes here, since the code I'm commenting on is already part of the codebase and I don't want to delay this further unnecessarily.

In the case of my above comments, you'd also want to run START_PROCESSING when a new magazine is loaded in, STOP_PROCESSING when it is ejected, and START_PROCESSING if the object is not already processing when fired.

Like I said, these changes aren't vital, and I'm marking as approved without them, but these sorts of measures are nice in making sure things aren't processing when they don't have to be - something I struggle with myself, at times.

PR #753, opened 2021-11-10
[READY FOR REVIEW] Adds Github's CI via Actions

Shadow-Quill:
# This is generally untested.
With so many files touched, it's fairly difficult for one person to test everything. I don't *believe* anything has been fucked, though a testmerge may be prudent.

## About The Pull Request

This PR *will* fail its own checks. Reasons being:
- Unformatted maps. This will be done in a later PR.
- Incorrect pixel offset in map(s). Same as above.
- Integration Tests will fail because discord.dm is failing. @vulppine was poked about this, and I was told he'll fix it Soon™.

also, any changes to core files were done upstream as well. (and I'm pretty sure I didn't do any custom edits myself)
- Note: I've slightly modified some sections so Integration Test logs are colored. They've been appropriately commented.
## Why It's Good For The Game

CI ensures that 2+2=4 in game (amongst other stuff, if a test has been written for it), and otherwise makes sure that a given PR can actually compile.

# Please merge this with the squash-type merge.

PR #769, opened 2021-12-01
Updates changelog workflows to only run on the main repo 

Shadow-Quill:
## About The Pull Request

Changelog actions shouldn't run on forks. Also lets the make_changelogs action actually run (as it was set to only run on Eris' repo)

### DIFF
--- a/.github/workflows/compile_changelogs.yml
+++ b/.github/workflows/compile_changelogs.yml
@@ -8,6 +8,7 @@ jobs:
   compile:
     name: "Compile changelogs"
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


Jamini:
The changelog workflows were nonfunctional anyway. Unless you changed this. They shouldn't have been running at all.

Shadow-Quill:
I updated the workflows as part of #753's workflow updates

Shadow-Quill:
As a note, a new repo secret will have to be created* with the name `CHANGELOG_ENABLER` and with any value.

*via the repo's Settings tab > Secrets > Actions > New Repository Secret

Jamini (approved):

PR #779, opened 2021-12-10
hot fix for siren scan event

RadiantFlash7:
Removes extranious var

## About The Pull Request
removes extra definition left over by sleep deprivation

## Why It's Good For The Game
Bug fix

## Changelog
```changelog
fix: fixes bug
```


Jamini (changes requested):
You can remove the callback ( the ..() ) from GiveTarget. It has no parent.

Jamini (approved):

PR #794, opened 2021-12-19
Comments projectile miss pending rework

RadiantFlash7:
## About The Pull Request
Comments out projectile miss temporarily.
I'll see if I can hassle another dev into fixing it so it won't miss pointblank/close range, and so that it doesn't hit things whiles missing them. Somehow.

## Why It's Good For The Game
Projectile miss is a good idea in theory to prevent people just aiming for head in PVP combat
Unfortunately, it bugs out PVE combat, and you can miss point blank with weapons...,
leading to funny situations. But also bad gameplay situations.

## Changelog
```changelog
code: comments out Projectile miss
```

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -254,8 +254,6 @@
 			else
 				def_zone = ran_zone(def_zone,spread)
 				result = target_mob.bullet_act(src, def_zone)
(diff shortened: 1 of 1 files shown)


Jamini (approved):
We need to actually fix this rather than just patch it, but approved for now.

PR #799, opened 2021-12-26
Readjusts artwork oddity stats as they were still coming up jank

Jamini:
## About The Pull Request

Adjusts how Artist and cube oddities are generated

## Why It's Good For The Game

Due to how I was doing this nerf before on the art oddities, we were losing points we should not have. That has been adjusted

Update!

Adjusted cube calculations to only round off fractional stats at the end of stat gen, instead of flooring both when calculating tool input stats as well as dividing by three. This should result in two power tools with 60 power generating a medium stat instead of a weak stat. ( For the final calculation we go from 4/3 (round to nearest int) to 4.8/3. Going from 1.33 round to nearest int (1) to 1.6 round to nearest int (2))

This brings cube oddities back in line to what they provided before (Medium oddity power with two power tools instead of weak)

## Changelog
```changelog
tweak: Artist oddities.
tweak: Cube oddities
```


Michiyamenotehifunana (approved):
Simple changes, all checks out.

PR #816, opened 2022-02-01
Unfuckers the deepmaint nuke

Shadow-Quill:
## About The Pull Request

Whenever the deepmaint `vault` chunk is selected, the nuke code in the ticker (which is handed out to joining heads) is overwritten, as deepmaint generates after the ship does.

Note: `eris_ship_bomb` should **_ONLY_** be set on the single nuke the heads are supposed to use.

## Why It's Good For The Game

Ship should be able to go boom.

## Changelog
```changelog
fix: Deepmaint nuke will no longer make the ship-side nuke un-armable without admin intervention.
```

### DIFF
--- a/maps/submaps/deepmaint_rooms/core/vault.dmm
+++ b/maps/submaps/deepmaint_rooms/core/vault.dmm
@@ -66,9 +66,8 @@
 /area/deepmaint)
 "aq" = (
 /obj/structure/railing{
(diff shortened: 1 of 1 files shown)


fludd12 (approved):
Double-checked with PR author to make sure the other map changes were intentional. They are notably _not_, but strongDMM just seems to have been trimming redundant map varedits.

PR #824, opened 2022-02-05
NVG Battery

Mocha-Not-Latte:
## About The Pull Request

Makes the Tactical Ballistic Helmet NVG consistent with the other NVG options. 
port of - https://github.com/discordia-space/CEV-Eris/pull/6686 

## Why It's Good For The Game

We like consistency don't we?

## Changelog
```changelog
tweak: Makes the Tactical Ballistic Helmet NVG now take batteries.
```

### DIFF
--- a/code/modules/clothing/glasses/night.dm
+++ b/code/modules/clothing/glasses/night.dm
@@ -15,14 +15,19 @@
 	. = ..()
 	overlay = global_hud.nvg
 
-/obj/item/clothing/glasses/bullet_proof_ironhammer
+/obj/item/clothing/glasses/powered/bullet_proof_ironhammer
 	name = "Night Vision Goggles"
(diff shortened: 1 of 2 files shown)


Shadow-Quill:
CI failing is due to some issue with `reader.dm`, which is not caused by this PR. For any commits **before** this comment, CI failure shouldn't block this PR.

Jamini (approved):
As per shadow. This looks fine.

Jamini (changes requested):
However as the path has changed, anything that uses the /obj/item/clothing/glasses/bullet_proof_ironhammer path needs to be updated to /obj/item/clothing/glasses/powered/bullet_proof_ironhammer path

Jamini (approved):
I am the dumb. The powered glasses are only referenced here. Should be good.

PR #828, opened 2022-02-10
Fixes "Get-Server-Log" not working if server hits midnight during a round

Shadow-Quill:
## About The Pull Request

Currently, if the server's time hits the next day, the server log for that round will no longer be accessible from in-game.
This makes it so that the "Get Server Log" verb uses the `diary` var, which is set to the server file for the round on world start, and doesn't change.

## Changelog
:cl:
admin: Get-Server-Log will no longer stop working if the server hits midnight during a round.
/:cl:

[And the upstream pr, as well.](https://github.com/discordia-space/CEV-Eris/pull/6998)


Jamini (approved):
Wonderful. You are a gem Shadow

PR #839, opened 2022-02-23
Fixes dmdocs

Shadow-Quill:
As per title, I think this will fix dmdocs not working properly.


Shadow-Quill:
I did not realize this would branch onto the main repo. Whoops.

Jamini (approved):

PR #854, opened 2022-03-19
Fixes a bug that corrupts a crafting step

Jamini:
## About The Pull Request

I need to test this PR yet. Please do not merge it.

This should fix an obscure bug where a tool flying away while you are doing a crafting step causes a turf to get passed to the crafting step. Corrupting that step for every single craft of an item for the remainder of the round.

This should provide some sanity checking to prevent that

## Why It's Good For The Game

Bugs bad

## Changelog
```changelog
fix: Crafting steps getting irrevocably corrupted
```

### DIFF
--- a/code/datums/craft/step.dm
+++ b/code/datums/craft/step.dm
@@ -152,6 +152,10 @@
 			return
 
 	else if(reqed_quality)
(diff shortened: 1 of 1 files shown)


Jamini:
Should be safe to merge. Tested and nothing seemed to break.

image

ShadzFen (approved):
Clean, commented. Nice

PR #857, opened 2022-03-26
Hotfix for allowing humans to spawn

Jamini:
We know the issue, but this should be a temporary fix until we can get the proper fix tested.

## About The Pull Request

Hotfix allowing humans to spawn

## Why It's Good For The Game

Something broke

## Changelog
```changelog
fix: Humans can now spawn
```


fludd12 (approved):
Approved for temporary merge. See developer chat for permanent solution.

PR #861, opened 2022-04-03
Medbay door changes

bearrrrrrrr:
## About The Pull Request

Adds a door-opening button to the surgery hallway to allow patients to leave without the doctors having to fully escort them out the surgery room, through the scanning room, into the trauma center. Additionally, removes the two doors leading from the scanning room to the trauma center.

Yellow: removal
Red: Addition

image

## Why It's Good For The Game

Better flow to medbay. Idc too much about the double doors but it's been a commonly-discussed thing by multiple people. When playing medical or filling a medical role, personally, it's a bit awkward to pass through them constantly- and ultimately, they take up space in a medbay that can get cramped awfully fast with multiple casualties, compacting with the fact you need to pass through them repeatedly, sometimes!

## Changelog
```changelog
add: 1 button to medbay
del: 2 doors from medbay
```

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -24192,8 +24192,8 @@
 /area/eris/maintenance/section3deck1central)
 "bfw" = (
 /obj/structure/table/standard,
-/obj/machinery/newscaster{
-	pixel_x = -32
+/obj/machinery/camera/network/medbay{
+	dir = 4
 	},
 /turf/simulated/floor/tiled/white/brown_perforated,
 /area/occulus/medical/surgeryhall)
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
However, to play devil's advocate, the set of double-doors does provide some amount of privacy for the more intimate parts of medical. This is also fulfilled by the left set of doors, but to a lesser extent.

In conclusion: idk i think it's worth at least trying for a bit and if the security and extra privacy is too precious then we add it

RadiantFlash7:
This was the original layout of the doors in medbay (minus the button). We ended up adding the two doors at the end of the scanner room to help with breaching, and as you said, for extra prviacy/security. 

I think the button addition would be helpful, but I’m less certain on reverting back to the old layout, personally.

bearrrrrrrr:
image
 discussion on discord had, we're good to go
 
 (by 'discussion' i mean 'other people discussed while i napped' poggers)

bearrrrrrrr:
> This is going to need to be completely redone due to the major update. Ergo why we are on freeze. You may want to wait until it's merged.

gotcha, will do

PR #879, opened 2022-05-07
fixes fishing hat stray pixel

anconfuzedrock:
whoops


anconfuzedrock:
also fixes orange hat being blue because it was set as blue but IT SHOULD BE ORANGE

Jamini (approved):

PR #891, opened 2022-07-02
add wedding dress

Awesome4Real:
## About The Pull Request

adds a wedding dress to the game & loadout menu. that's it.

## Why It's Good For The Game

for the event on Sunday. (this is only meant to be temporary)

## Changelog

adds a wedding dress to the game & loadout menu.


fludd12 (approved):
Everything looks good to me! You don't need to make it temporary, though - I think keeping this in the code is a good idea.

fludd12:
As a side note, the failing check is the issues with our Oldificator linter working on randomly generated loot, and attempting to Oldify one of the few loots that has no reaction to the proc. Re-running the check one or more times will clear it out.

PR #894, opened 2022-08-02
Caseless Ammo Fix

Awesome4Real:
## About The Pull Request

Makes caseless ammo not drop casings when the ammo was previously pulled out of a magazine/ammo box and put back into a magazine.

## Why It's Good For The Game

Not having those casing errors everywhere.

## Changelog
tweak: made the function for taking ammo out of a magazine/putting it in copy over the "is_caseless" variable to the new bullets.
```

<!-- Leave the codeblock and the "changelog" alone for your PR to have working automatic change-log generation. -->
<!-- You can use multiple of the same prefix (they're only used for the icon ingame) and delete the unneeded ones. Despite some of the tags, changelogs should generally represent how a player might be affected by the changes rather than a summary of the PR's contents. -->

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -56,6 +56,7 @@
 		new_casing.icon_state = src.icon_state
 		new_casing.spent_icon = src.spent_icon
 		new_casing.maxamount = src.maxamount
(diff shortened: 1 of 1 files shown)


Jamini (approved):
I am so glad someone fixed this.

PR #897, opened 2022-08-07
Silkcraft fixes

Zenitheevee:
## About The Pull Request

Silk requirement for webs lowered to 1 (Its going to be a PITA to get enough to really use them if its 3, and so i dont have to fuck around with more checks and questions about "why do you need 3 to make webbing when you're harvesting it all to just 1 glob?") and added more checks for webs to avoid stacking
Silk covers recipe seperated from web covers

## Why It's Good For The Game

Because its a PITA

## Changelog
```changelog
balance: Silk globs have a cost of 1 for making webs now 
fix: web covers aren't over-written
```


fludd12 (approved):
Looks good! It's good that you added the no doubling-up on web thing, too, as that can be pretty nuts.

PR #900, opened 2022-08-16
Fixes synthetics sending soulcrypts to the shadow realm

Jamini:
## About The Pull Request

The soulcrypt reader in NT cloning now drops the crypt on the floor instead of your hand.

## Why It's Good For The Game

This prevents synthetics from sending the core implant into the shadow realm. (an inaccessible part of the core implant reader)

## Changelog
```changelog
fix: Core implant readers no longer make soulcrypts inaccessible if a synthetic touches the device.
```

### DIFF
--- a/code/modules/core_implant/cruciform/machinery/cloning.dm
+++ b/code/modules/core_implant/cruciform/machinery/cloning.dm
@@ -502,10 +502,10 @@
 		return
 
 	if(reading)
-		to_chat(user, SPAN_WARNING("You try to pull the [implant], but it does not move."))
+		to_chat(user, SPAN_WARNING("You try to eject the [implant], but it does not move.")) //Occulus Edit: Synthetics no longer delete crypts
 		return
 
-	user.put_in_active_hand(implant)
+	implant.forceMove(loc) //Occulus Edit: Synthetics no longer delete crypts
 	implant = null
 
 	src.add_fingerprint(user)


Shadow-Quill (changes requested):
If you add a `put_in_hands` proc after the force move (may need an `!issilicon` check to prevent runtimes), you can have the implant still be put into human hands without losing it to the aether.

Jamini:
I don't really see the point in doing so. Originally I had made it more complex and check the type of the entity using the device, but ultimately KISS actually would apply here.

There just isn't a good reason to put this object directly into someone's hands, and it can cause other bugs (such as swapping hands too quickly on a humanmob causing the object to get lost. Which is currently an issue with chemistry equipment.)

Shadow-Quill:
It's quality of life, if anything, and the put in hands proc only has to be in a `if(!issilicon)` check

Jamini:
> Actually- I've just checked, and `put_in_hands()` also handles if it can't be put in the mob's hands (seemingly if an AI ejects it), and if it can't be put in the hands, it's just ejected onto the floor.
> 
> https://github.com/Occulus-Server/Occulus-Eris/blob/1f19288f7154109940ba298d5dee38b09c161295/code/modules/mob/inventory.dm#L39-L49

This suggested change causes soulcrypts to teleport to the AI core (or to a remote borg/drone if they aren't directly adjacent.)

Jamini:
Screenshot_20220817_110813

fludd12 (approved):
Looks good, as long as you're alright slapping soulcrypts onto the floor. Which honestly isn't a huge problem.

PR #917, opened 2022-09-27
Custom blood colors

gozulio:
## About The Pull Request

Gives players the ability to set a custom blood color in their character options.
Also fixes a bug with the Medical Stand when filling blood bags.

## Why It's Good For The Game

Commonly requested feature by players

## Changelog
```changelog
add: Custom character blood colors
fix: Fixed Medical stand making several hundred unit long reagent tables of the same kind of blood
fix: Runtime when admin healing a human mob with no client
```


code/game/objects/structures/medical_stand.dm
```diff
@@ -436,7 +436,7 @@
 
 			var/datum/reagent/B = H.take_blood(beaker,amount)
 			if (B)
-				beaker.reagents.reagent_list |= B
+				beaker.reagents.add_reagent(B.id, B.volume, B.data)  //Oculous Edit - Fix obnoxious IV reagent list bug
```
Shadow-Quill, line 439:
```suggestion
				beaker.reagents.add_reagent(B.id, B.volume, B.data)  //Occulus Edit - Fix obnoxious IV reagent list bug
```

zzzz_modular_occulus/code/modules/client/preference_setup/loadout/02_body.dm
```diff
@@ -1,3 +1,14 @@
+datum/preferences
```
Shadow-Quill, line 1:
```suggestion
/datum/preferences
```

Shadow-Quill (changes requested):

zzzz_modular_occulus/code/modules/client/preferences.dm
```diff
@@ -95,6 +95,7 @@
 	character.appearance_flags	-= fuzzy*PIXEL_SCALE
 
 	character.body_markings = body_markings
+	character.species.blood_color = blood_color
```
Michiyamenotehifunana, line 98: Every subsequent player that spawns in will override the blood color of any existing players in-round who has the same species, as this directly modifies the species datum shared by everyone who selects that particular species.

Michiyamenotehifunana (changes requested):

zzzz_modular_occulus/code/modules/client/preferences.dm
gozulio: This is correct. I'm working on a fix for it now.

gozulio: There is two possible solutions. One is to try and make the player mobs species mob it's own local datum rather then referencing a global one.
The other option is to move the blood color onto the mob itself, and then change all the blood decal code to reference the mobs blood color instead of the mobs species blood color.

As you can tell with the commits I just made I've opted for the latter. However I realize in the past this had a problem where players would constantly fill their reagent tables with blood that wasn't theirs. So far this has not been the case in my testing but any relevant information on how I can test for this would be helpful.

Shadow-Quill (changes requested):

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1136,10 +1136,12 @@
 	sanity.setLevel(sanity.max_level)
 	restore_blood()
 //OCCULUS EDIT - Resets bodymarkings when we rejuv
+	if(!client)
+		..()
+		return
```
Shadow-Quill, line 1141: and this block should prolly go under the `for(var/N in organs_by_name)` block

Shadow-Quill (changes requested):

code/modules/mob/living/carbon/human/life.dm
gozulio: Then we'd just end up cutting markings from mobs that most likely don't have any.

Or if our mob is a player mob without a ghost then we cut their markings without adding them back.

I'm pretty sure markings aren't interacted with in any way outside of character creation.

Shadow-Quill: ah, I didn't see that L1142-1144 was removing markings, I thought it was replacing organs. Whoops.

Shadow-Quill (approved):

PR #923, opened 2022-11-17
Clumsily Adds A Rudementary Implant

Ikmachek:
## About The Pull Request
Adds dummy implants and an implanter to the loadout almost like Virgo has. I did my best to get them to just place within the mob like they also have but that requires a tgui element we'll never have, so my next best idea was toss in an implanter. If someone smarter than me wants to take a crack at this, by all means. However the implants can be named, retain their names and descs, so it's good enough for me!

## Why It's Good For The Game
Silly little flavor to push the idea minor cyber and bio augments are part of the setting.

## Changelog
```changelog
add: Custom implants/augments ala Virgo.
```

### DIFF
--- a/zzzz_modular_occulus/code/modules/loadout/lists/misc.dm
+++ b/zzzz_modular_occulus/code/modules/loadout/lists/misc.dm
@@ -36,23 +36,17 @@
 	)
 	gear_tweaks += new /datum/gear_tweak/path(balls)
 
-/datum/gear/implant/generic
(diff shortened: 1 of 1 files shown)


Trilbyspaceclone:
The reason why its failing to past the integration tests is do to all those being a subtype of
"/datum/gear/implant"
that has no values such as display name.

Shadow-Quill (approved):

PR #925, opened 2022-11-30
TGUI-ification

Shadow-Quill:
## About The Pull Request
Adds TGUI.

## Why It's Good For The Game

ohmygod it's such a better interface to use.
also clobbers in a whole lotta changes from upstream, like asset caching, master controller updates, and other stuff.

absolutely this **MUST** be testmerged before merging

NanoUI still exists everywhere except for the changelog panel, as that is the only interface which Eris transferred over.

## Changelog
```changelog
add: Added TGUI.
```


Shadow-Quill:
this shoooould be able to be testmerged now. the failing check is due to a map apc check integration test failing for some reason.

ShadzFen:
A-SFM-GoingForPizza

PR #930, opened 2022-12-21
Mark-Recapture Roaches

Takykardi:
## About The Pull Request

Adds two "new" types of roaches. As in, just normal roachlings and kampfers with pink paint on their butts. This is in response to a prompt after discussion about the results of an in character experiment that was performed.

## Why It's Good For The Game

In character, an experiment was begun to estimate the population size of the roaches aboard the Light using a standard mark-recapture methodology. This... did not go very well. Within hours, even freshly hatched roachlings began to show the pink paint.
And a thanks to Fel for helping me put the var/spawn_type in the 'right' place outside of the if blocks so that way it's A. less of a fustercluck visually and B. doesn't need every single one of them to have that var/ bit.

## Changelog
```changelog
add: code for the pink butt roaches, about a 1 in 50 chance for them to show up
imageadd: new pink butt roaches
code: put the var/spawn_type in roaches.dm in the right place
```


Shadow-Quill (approved):

Jamini:
The madlad actually did it.

PR #941, opened 2023-01-19
No law zero'd drones

Shadow-Quill:
## About The Pull Request

Maintenence drones shouldn't be forcefully law zero'd when an AI becomes malf.
The only time they should get those laws is when emagged.

## Changelog
```changelog
fix: Maintenence drones will no longer be force synced to a AI if it becomes an antag.
```


Jamini (dismissed):

Jamini:
This is failing to compile

Jamini (approved):

PR #943, opened 2023-01-20
Accessory Update

Mocha-Not-Latte:
## About The Pull Request
Adds in the ability to put accessories on your armor, and a decent amount of new options. Aswell as the ability to change the CAS Harness to an armored vest

## Why It's Good For The Game
Cosmetics :)


zzzz_modular_occulus/code/modules/clothing/masks/masks.dm
```diff
@@ -2,70 +2,3 @@
 	body_parts_covered = 0
 	style = STYLE_HIGH	// Scarves oughta be stylish!
 
-/obj/item/clothing/mask/scarf/shemaugh
```
Jamini, line 5: This means they will lose the style bonus, which is something of a shame.

Jamini (changes requested):
Please move the modular changes to core files to a modular file. Nothing that was changed needs to be inline here.

Jamini (approved):
exciting stuff

PR #961, opened 2023-04-19
new encounter map: one-star biofacility

bearrrrrrrr:
## About The Pull Request

bio-facility onestar map. probably going to be an overmap encounter after being used as an event map

## Changelog
```changelog
add: start of one-star bio-facility
```


bearrrrrrrr:
this motherfucker

Jamini (approved):

PR #983, opened 2023-05-18
Siren Weight Adjust & Required Crew Fix

WeNeedMorePhoron:
## About The Pull Request
- Adjust siren weight to 0.25 and required crewmember to 12 (Instead of the intended 8 which ended up as a de facto 8). 
- Remove the max_crew_diff_lower variable which defaulted to 10, leading to all event without it set explicitly having 10 lower required crew than expected.

## Why It's Good For The Game
Make Siren no longer fire on low pop, lower its chance of firing. Also align coder expectations of req_crew blocking events firing on low pop round with what is happening. 

CHANGES: 
Because of these changes, numerous event's actual required crew has been adjusted upward. I am open to feedback to changing it back.
- Malf AI now fire on min. 15 pop instead of 10 pop (might need adjustment?)
- Marshal fire on min. 10 pop instead of 0 pop
- Mercenary fire on min. 5 pop instead of -5 pop
- Siren fire on min. 12 pop instead of 5 pop

## Changelog
```changelog
tweak: Siren have half the initial weight and can only fire when there's 12 crew or more 
tweak: Malf AI now fire on minimal 15 pop instead of 5 pop. 
tweak: Marshal now fire on minimum 10 pop and mercenary 5 instead of 0 / -5 which was not working as intended. (These events can fire but fail to find a target)
```

### DIFF
--- a/code/game/gamemodes/storyevent.dm
+++ b/code/game/gamemodes/storyevent.dm
@@ -32,7 +32,8 @@
 
 	var/req_stage = -1
 
-	var/max_crew_diff_lower = 10	//Maximum difference between above values and real crew distribution. If difference is greater, weight will be 0
+	// var/max_crew_diff_lower = 10		// Occulus Edit: Removed since this is no longer used as the original
+	// And it was causing required crew to be unexpectedly low (reduced by 10) on all events, leading to coder confusion and unintentional event firing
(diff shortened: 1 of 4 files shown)


Jamini:
Your changes do not do what you think they do, and this removes the scaling percentage below ideal population.

PR #989, opened 2023-05-20
Add stomach pump to sleeper

WeNeedMorePhoron:
## About The Pull Request
Add Stomach Pump to sleeper, which remove 10 (out of 400 max possible) nutrition per tick and also remove stomach (Ingested) reagents at the same rate as Dialysis. 

Great when your cat eat bad maint food.

## Why It's Good For The Game
gives a reliable way to deal with reagents in stomach that isn't potassium water pill or ingesting two dispenser supply of carbon (Exaggerration)

i'm feeling real pumped for this change

## Changelog
```changelog
add: Stomach Pump added to sleeper. Works similarly to dialysis but clear out your stomach's content. Will nuke your nutrition on use (10 / 400 nutrition removed per tick).
```

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -39,17 +39,17 @@
 
 	// Occulus Edit: Stomach Pump
 	if(pumping)
-		if(!beaker) return
+		if(!beaker)
+			toggle_pump()
 
 		if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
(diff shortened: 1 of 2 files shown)


Shadow-Quill (changes requested):

WeNeedMorePhoron:
Updated. Reused the pumped variable to scale the nutrition drain. Cleaned up the code and tested it works as intended.

Shadow-Quill (approved):

PR #999, opened 2023-05-23
Cleans up Circuit disk addition + Re-allows crafting station printing

Zenitheevee:
## About The Pull Request

Replaces a hard over-write with an initialization addition to the list

## Why It's Good For The Game

hard overwrites bad, re-enables obtainment for crafting station

## Changelog
```changelog
fix: Replaces the circuit disk overwrites with a better addition
```

### DIFF
--- a/zzzz_modular_occulus/code/game/objects/items/weapons/design_disks/ees.dm
+++ b/zzzz_modular_occulus/code/game/objects/items/weapons/design_disks/ees.dm
@@ -30,28 +30,13 @@
 
 /obj/item/computer_hardware/hard_drive/portable/design/circuits
 	disk_name = "EES-ESPO-830 Circuits"
-	designs = list(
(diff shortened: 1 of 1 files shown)


WeNeedMorePhoron (approved):

Zenitheevee:
> Why are we doing it like this?
> 
> We use hard-overwrites to avoid a lot of issues with init setup and potential runtimes.

Just parroting how we used the method for adding like we did on the rubber disk, even before the recent relocation and addition

And this system is more friendly for porting changes. If its a big problem i can just go back to a hard overwrite.

Jamini:
Our modular files overwriting is very intentional however. Quite a lot of upstream "Changes" are intentionally reverted or blocked with modular overwrites.

I'd prefer we kept it as it was if that's alright.

Shadow-Quill (approved):

PR #1010, opened 2023-06-08
Breakdown Rework Part 1: Simple Progressive Breakdown

WeNeedMorePhoron:
## About The Pull Request
This PR is meant to be an intermediate solution to complaints about the breakdown system. Numbers subject to feedbakc.

Proof of Concept for Progressive Breakdown, part 1. There were a lot of suggestions discussed in the thread and this is what I can reasonably get out in a day. Eventually this should be developed into a more elaborate system.

Part two is implementing the Jaded Perk.

- Negative breakdown probability no longer matter. Kept for upstream compatibility 
- Your first breakdown will always be a positive or common (minor) breakdown 
- If you roll a breakdown without objective, you're lucky and nothing follows 
- If you roll a common breakdown with an objective, and fail to fulfill it, your next breakdown in an hour is guaranteed to be a negative breakdown 
- If you fulfill the objective your next breakdown will once again be a normal / positive breakdown

Also add Forgotten breakdown, a mirrored counterpart to A Lesson Learnt, as step 1 to add in more breakdowns and move away from breakdown that totally take away player agency. 

## Why It's Good For The Game
- Restore player agency by giving player forewarning before they get a breakdown that robs them of agency. 
- Increases the predictability of the system by making the harsh punishment a direct, and unavoidable consequences of neglecting sanity for a long period. 
- Encourage interaction with the system - meaning players have to engage with objective breakdown instead of ignoring it and get lucky until they don't

## Changelog
```changelog
add: A new forgotten lesson breakdown that hits your stat across the board between 2 to 5, a (weaker) mirror to "A Lesson Learnt" positive breakdown.
tweak: Tweaked the breakdown system. Now you cannot roll a negative breakdown on first try. Instead, negative breakdown occurs as a result of getting a normal breakdown with objective but failing to fulfill the objective. If you don't get any breakdown with objective, there's nothing you need to do.
```

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -3213,6 +3213,7 @@
 #include "zzzz_modular_occulus\code\modules\research\nodes\engineering.dm"
 #include "zzzz_modular_occulus\code\modules\research\nodes\illegal.dm"
 #include "zzzz_modular_occulus\code\modules\research\xenoarchaeology\tools\anomaly_suit.dm"
+#include "zzzz_modular_occulus\code\modules\sanity\breakdowns.dm"
 #include "zzzz_modular_occulus\code\modules\sanity\sanity_mob.dm"
 #include "zzzz_modular_occulus\code\modules\scrap\oldificator.dm"
 #include "zzzz_modular_occulus\code\modules\shield_generators\shield_generator.dm"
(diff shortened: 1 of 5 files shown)


WeNeedMorePhoron:
I chatjeepityed those phrases and I'm pretty sure they are way too long so someone with an English major give some suggestions ty

Jamini (approved):

PR #1011, opened 2023-06-09
Remove some price_tag that broke economy, adjust organ price tag to 2…

WeNeedMorePhoron:
## About The Pull Request
Nuked part of the price_list file which no longer exists upstream - namely the part that defined water as 20 per unit and alcohol as 50 - 250 per unit. It is a bit ridiculous.

Also brought organ price in line with upstream to 200 instead of 500.

## Why It's Good For The Game
Prevents me from gaming the cargo system too hard until rebase come. This is way too tempting, y'know.

## Changelog
```changelog
tweak: Water no longer sell for 20 credit per unit, alcohol no longer sell for 50 - 250 credit per unit (they are mostly worthless out of container now).
tweak: Organ sell for 200 instead of 500 as a baseline, in line with upstream.
```


Jamini:
We already fixed water and alcohol prices modularily iirc

PR #1016, opened 2023-06-14
Traitor item: Portable Autodoc

Awesome4Real:
## About The Pull Request

This PR adds the Portable Autodoc to the traitor's uplink. This device costs 20 tc, and doubles as both a place-able costless autodoc, as well as a defib while worn on the back.

image
image

## Why It's Good For The Game

Right now, the only way for antagonists like contractors or lone mercs to preform self-surgery is to purchase a rigsuit, then purchase the autodoc upgrade for it. This is expensive and not obvious, especially for new players. Also, there is currently no way for mercs to revive their fallen allies if they die. This PR kills two birds with one stone.

Of course, this autodoc is expensive in tc, is obvious to anyone who sees it, and takes a long time to work. You will need to find a secure area in order to finish the procedures without being caught.

## Changelog
```changelog
add: Portable Autodoc
```


WeNeedMorePhoron (approved):
Adding an examine text that tells player to "use" the item for deploying it would be nice

Awesome4Real:
i made that change, thanks for the recommendation!

WeNeedMorePhoron (approved):

PR #1038, opened 2023-07-13
removes the turret from chromelab.dmm

bearrrrrrrr:
## About The Pull Request

removes the turret from chromelab.dmm since it never actually worked, replaces it with a gunba.

## Why It's Good For The Game

eh. didn't feel right to add a turret control panel in a randomly-generating place, esp when more than one turret might random gen from other rooms. also adds a bit more loot in it since it was kinda loot-sparse imo

## Changelog
```changelog
del: removes the turret from chromelab.dmm
```


bearrrrrrrr:
image

WeNeedMorePhoron (approved):

PR #1048, opened 2023-07-16
Fix CI config, again

martinlyra:
## About The Pull Request

I missed an option for the copy command run in the CI. Didn't catch it because the CI was known to be broken for some while and I just neglected it.

## Why It's Good For The Game

CI, please work. We need you.

## Changelog
```changelog

```

### DIFF
--- a/.github/workflows/ci_suite.yml
+++ b/.github/workflows/ci_suite.yml
@@ -73,9 +73,6 @@ jobs:
         with:
           path: $HOME/BYOND
           key: ${{ runner.os }}-byond
-      - name: Setup example config
-        run: |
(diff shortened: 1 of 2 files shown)


martinlyra:
It fixes the error, but it does not fix the config files not being copied over to the config for some reason

martinlyra:
Well, now the configuration issue has been finally resolved. But it does not fix the server attempting to use the SQL for loading admins, and it is not working out.

Unfortunately I am the kind of person who hates SQL with genuine passion, I am not going anywhere that.

Jamini (approved):

PR #1050, opened 2023-07-19
Revert "Revert "Fix CI config, again""

martinlyra:
Reverts Occulus-Server/Occulus-Eris#1049

I am personally confident in my fix, but better make it twice to be sure if nothing is crazy.


Jamini (approved):
Approving. Let's get this all working

PR #1058, opened 2023-09-04
Update ADMS.dm

Jamini:
Adds the explore portions of the ADMS to the list at even lower rates. Because people were mistakenly thinking that the ADMS doesn't work on the asteroid.

## About The Pull Request

Further expands zones the ADMS functions in

## Why It's Good For The Game

More science good

## Changelog
```changelog
tweak: ADMS now also gives a very slow point rate while on the asteroid
```


fludd12 (approved):
Approved! Silly, though. We should totally do some fancy stuff at some point for the ADMS.

